PR #9, opened 2025-09-23
[RTOP-69] implement base code to commands

lucasbutzke:
1. run_compile() Function - Program Compilation

Compiles a program and manages the entire build process.

What it does:

Starts compilation: Executes a bash script (compile.sh)
Monitors progress: Shows real-time output and error messages
Manages build state: Tracks compiling/success/failed status
Handles system restart: Stops and starts the PLC controller after compilation
Performs cleanup: Runs a cleanup script (compile-clean.sh)
Step-by-step flow:

Sets status to "COMPILING"
Starts the compilation script
Creates threads to monitor process output
Waits for compilation to complete
If successful: stops PLC, runs cleanup, restarts PLC
If failed: shows error message

2. handle_upload_file() Function - File Upload Handling

Receives a ZIP file, validates it, and extracts contents.

What it does:

Validations: Checks file safety (prevents zip bombs)
Preparation: Cleans old directory and extracts new ZIP
Initiates compilation: Calls compilation function in separate thread
Validation checks:

✅ Verifies no active compilation is running
✅ Confirms file exists in request
✅ Checks file size limits
✅ Analyzes ZIP safety
✅ Extracts only valid files

Complete Flow:
Upload ZIP → Validate → Clean folder → Extract → Compile → Stop PLC → Move file → Start PLC


core/src/plc_app/unix_socket.c
```diff
@@ -39,57 +39,55 @@ static ssize_t read_line(int fd, char *buffer, size_t max_length)
     return total_read;
 }
 
-void handle_unix_socket_commands(char *command, char *response, size_t response_size)
+void handle_unix_socket_commands(const char *command, char *response, size_t response_size)
 {
-    if (strcmp(command, "PING") == 0)
-    {
+    if (strcmp(command, "PING") == 0) {
```
marconetsf: Most of the functions and ifs, elses etc... are not in the clang-format default. (accordingly with BARR standard)

``` C
void function()
{
// code
}
```

instead of

``` C
void function() {
// code
}
```

PR #32, opened 2025-11-12
Rtop 84 master plugin tests

lucasbutzke:
Pytest for Modbus plugin


core/src/drivers/plugins/python/modbus_master/modbus_master_plugin.py
```diff
@@ -59,617 +59,6 @@ def get_batch_write_requests_from_io_points(io_points: List[Any]) -> Dict[int, L
 runtime_args = None
 modbus_master_config: ModbusMasterConfig = None
 safe_buffer_accessor: SafeBufferAccess = None
-slave_threads: List[threading.Thread] = []
```
marconetsf, line 62: Where all the ModbusSlaveDevice class implementation went?

marconetsf (approved):
LGTM

PR #80, opened 2026-01-19
Add Linux gpiod plugin

QB4-dev:
I have created a plugin to support generic Linux gpio using libgpiod(the libgpiod-dev package is required to build). It should provide OpenPLC gpio control on any Linux device with GPIOs. I think it is better way than using hardware dependent libraries like wiringPi

To get it work user should provide IEC address to gpio csv map file like:
**File: io-map.csv**

```csv
IEC_ADDRESS,CHIP_NAME,LINE_IDENTIFIER
%IX0.2,/dev/gpiochip0,GPIO2
%IX0.3,/dev/gpiochip0,GPIO3
%IX0.4,/dev/gpiochip0,GPIO4
%IX0.3,/dev/gpiochip0,GPIO5
%QX0.0,/dev/gpiochip0,GPIO10
%QX0.1,/dev/gpiochip0,GPIO11
%QX0.2,/dev/gpiochip0,GPIO12
%QX0.3,/dev/gpiochip0,GPIO13
```
And then activate plugin in plugins.conf 

GPIOs from map should be controlled by OpenPLC

### DIFF
--- a/core/src/drivers/plugins/native/gpiod/Makefile
+++ b/core/src/drivers/plugins/native/gpiod/Makefile
@@ -1,0 +1,28 @@
+# Makefile for test plugin
+
+CC = gcc
+CFLAGS = -fPIC -I../../../../lib -I../../../../drivers -I..
+LDFLAGS = -shared -lpthread -lgpiod
+
+TARGET = gpiod_plugin.so
+SOURCE = gpiod_plugin.c
+PLUGIN_LOGGER = ../plugin_logger.c
+
+LOADER_TARGET = test_plugin_loader
+LOADER_SOURCE = test_plugin_loader.c
+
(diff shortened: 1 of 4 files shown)


thiagoralves:
Thank you very much for this PR. This is very interesting, but we might put it on hold for a little bit until we figure out a good way to interface it from the editor in a user friendly way.

QB4-dev:
Okay. I stuck with tests just because program generated from editor disables my plugin. I see that it is all work in progress. 

I'm Linux embedded developer. For my point of view it will be good to create Yocto recipe to build runtime and then supply correct IEC -> GPIO maps depending on target device(machine in Yocto). There is no need to configure I/O from editor

BTW Thank you for your great work

PR #114, opened 2026-03-05
feat: Three-layer crash recovery to prevent restart loops

thiagoralves:
## Summary

- **Layer 1 - Signal handler**: Catch SIGFPE/SIGSEGV in the PLC cycle thread via `sigsetjmp`/`siglongjmp`, transitioning to ERROR state while keeping the runtime process alive so the command socket remains available for uploading a corrected program
- **Layer 2 - Watchdog recovery**: Transition to ERROR state instead of `exit()` when the watchdog detects an unresponsive PLC (e.g. infinite loop), keeping the runtime alive for recovery
- **Layer 3 - Webserver safe mode fallback**: Track crash frequency with a sliding window; after 3 crashes within 30 seconds, restart the runtime with `--safe-mode` flag which skips loading the PLC program, allowing a new program to be uploaded
- Reset crash tracking state after successful program compilation

## Context

When a PLC program contains fatal errors (division by zero, bad array access, infinite loops), the runtime would crash and the webserver would blindly restart it in an infinite loop, making recovery impossible without manually killing the processes.

## Test plan

- [ ] Upload a PLC program with division by zero (SIGFPE) — runtime should enter ERROR state, process stays alive, new program can be uploaded
- [ ] Upload a PLC program with out-of-bounds array access (SIGSEGV) — same recovery behavior
- [ ] Upload a PLC program with an infinite loop — watchdog transitions to ERROR, runtime stays alive
- [ ] Kill the runtime process manually 3 times rapidly — webserver should detect crash loop and restart in safe mode
- [ ] After safe mode, upload a valid program — crash tracking resets, PLC runs normally
- [ ] Normal PLC programs continue to work without any behavioral changes

### DIFF
--- a/core/src/plc_app/plc_main.c
+++ b/core/src/plc_app/plc_main.c
@@ -36,6 +36,7 @@ void handle_sigint(int sig)
 int main(int argc, char *argv[])
 {
     bool print_debug = false;
+    bool safe_mode   = false;
 
     // Check for command line arguments
     for (int i = 1; i < argc; i++)
@@ -48,11 +49,9 @@ int main(int argc, char *argv[])
         {
             print_debug = true;
         }
-
-        // Early exit if both flags are found
-        if (print_logs && print_debug)
+        else if (strcmp(argv[i], "--safe-mode") == 0)
         {
-            break;
+            safe_mode = true;
         }
     }
 
@@ -98,8 +97,14 @@ int main(int argc, char *argv[])
         return -1;
     }
 
-    // Start PLC
-    if (plc_set_state(PLC_STATE_RUNNING) != true)
+    // Start PLC (skip in safe mode to allow program upload without loading the
+    // faulty program that may have caused repeated crashes)
+    if (safe_mode)
+    {
+        log_info("Runtime started in SAFE MODE - PLC program will not be loaded");
+        log_info("Upload a corrected program to recover");
+    }
+    else if (plc_set_state(PLC_STATE_RUNNING) != true)
     {
         log_error("Failed to set PLC state to RUNNING");
     }
--- a/core/src/plc_app/plc_state_manager.c
+++ b/core/src/plc_app/plc_state_manager.c
@@ -28,6 +28,7 @@ extern plugin_driver_t *plugin_driver;
 static sigjmp_buf plc_crash_jmp;
 static pthread_t plc_thread_id;
 static volatile sig_atomic_t plc_crash_signal = 0;
+static volatile sig_atomic_t holding_buffer_mutex = 0;
 
 static void plc_crash_handler(int sig)
 {
@@ -50,6 +51,7 @@ void *plc_cycle_thread(void *arg)
 
     // Record this thread's ID for the crash handler
     plc_thread_id = pthread_self();
+    plc_crash_signal = 0;
 
     // Initialize PLC with real-time optimizations
     set_realtime_priority();
@@ -122,8 +124,12 @@ void *plc_cycle_thread(void *arg)
     if (crash_sig != 0)
     {
         // We got here via siglongjmp from the crash handler.
-        // Release the buffer mutex in case we held it when we crashed.
-        plugin_mutex_give(&plugin_driver->buffer_mutex);
+        // Only release the buffer mutex if we held it when we crashed.
+        if (holding_buffer_mutex)
+        {
+            holding_buffer_mutex = 0;
+            plugin_mutex_give(&plugin_driver->buffer_mutex);
+        }
 
         const char *sig_name = (crash_sig == SIGFPE) ? "SIGFPE (arithmetic error, e.g. division by zero)"
                                                       : "SIGSEGV (memory access violation)";
@@ -147,6 +153,7 @@ void *plc_cycle_thread(void *arg)
     while (plc_state == PLC_STATE_RUNNING)
     {
         scan_cycle_time_start();
+        holding_buffer_mutex = 1;
         plugin_mutex_take(&plugin_driver->buffer_mutex);
 
         // Apply pending journal entries before plugin hooks run
@@ -167,6 +174,7 @@ void *plc_cycle_thread(void *arg)
         atomic_store(&plc_heartbeat, time(NULL));
 
         plugin_mutex_give(&plugin_driver->buffer_mutex);
+        holding_buffer_mutex = 0;
         scan_cycle_time_end();
 
         // Calculate next start time
@@ -392,6 +400,14 @@ void plc_state_manager_cleanup(void)
     {
         unload_plc_program(plc_program);
     }
+}
+
+void plc_force_error_state(void)
+{
+    pthread_mutex_lock(&state_mutex);
+    plc_state = PLC_STATE_ERROR;
+    pthread_mutex_unlock(&state_mutex);
+    log_info("PLC State: ERROR");
 }
 
 int plc_get_crash_signal(void)
--- a/core/src/plc_app/plc_state_manager.c
+++ b/core/src/plc_app/plc_state_manager.c
@@ -1,5 +1,8 @@
 #include <pthread.h>
+#include <setjmp.h>
+#include <signal.h>
 #include <stdatomic.h>
+#include <string.h>
 
 #include "../drivers/plugin_driver.h"
 #include "image_tables.h"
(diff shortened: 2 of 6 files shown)


core/src/plc_app/plc_state_manager.c
```diff
@@ -21,10 +24,33 @@ extern plc_timing_stats_t plc_timing_stats;
...
+static void plc_crash_handler(int sig)
+{
+    // Only handle if the crash came from the PLC cycle thread
+    if (!pthread_equal(pthread_self(), plc_thread_id))
+    {
+        // Not our thread - restore default handler and re-raise
+        signal(sig, SIG_DFL);
+        raise(sig);
+        return;
+    }
```
thiagoralves, line 56: **Concern: `siglongjmp` from a hardware-raised signal handler**

While `siglongjmp` is technically async-signal-safe per POSIX, jumping out of a SIGSEGV handler is only well-defined if the signal was caused by `raise()` or `kill()`, not by actual invalid memory access. When SIGSEGV comes from corrupted heap/stack state, `siglongjmp` may land in an inconsistent state (e.g., heap allocator locks held, corrupted stack frames).

This is a known limitation of the approach and probably acceptable given the alternatives, but it would be worth adding a comment here documenting that after a SIGSEGV recovery, the process state may be unreliable and Layer 3 (safe mode restart) exists as the backstop for that case.

thiagoralves (commented):
Overall the three-layer design is well thought out and addresses a real usability problem. Each layer catches a different failure mode cleanly. A few concerns below, mostly around signal safety and concurrency edge cases.

core/src/plc_app/plc_state_manager.c
```diff
@@ -75,6 +115,35 @@ void *plc_cycle_thread(void *arg)
...
 
+    // Set up the crash recovery point. sigsetjmp returns 0 on initial call,
+    // and returns the signal number when siglongjmp jumps back here after
+    // a crash in the PLC program.
+    int crash_sig = sigsetjmp(plc_crash_jmp, 1);
+    if (crash_sig != 0)
+    {
+        // We got here via siglongjmp from the crash handler.
+        // Release the buffer mutex in case we held it when we crashed.
+        plugin_mutex_give(&plugin_driver->buffer_mutex);
```
thiagoralves: **Bug: unconditional mutex release may be undefined behavior**

If `buffer_mutex` was not held at the time of the crash, calling `plugin_mutex_give` here is an unlock of a mutex not owned by the calling thread -- behavior is undefined per POSIX pthreads. Depending on the mutex type, this could corrupt the mutex state or be silently ignored.

Consider either:
- Using `pthread_mutex_trylock` followed by `pthread_mutex_unlock` (trylock will succeed if it wasn't held, then unlock releases it; if it was held by us, trylock fails with `EDEADLK` for error-checking mutexes or we can track ownership with a flag)
- Setting a `volatile` flag before/after the lock in the main loop and checking it here

core/src/plc_app/plc_state_manager.c
```diff
@@ -21,10 +24,33 @@ extern plc_timing_stats_t plc_timing_stats;
 extern atomic_long plc_heartbeat;
 extern plugin_driver_t *plugin_driver;
 
+// Signal recovery for PLC cycle thread crashes (SIGFPE, SIGSEGV)
+static sigjmp_buf plc_crash_jmp;
```
thiagoralves, line 28: **Minor: `plc_crash_signal` is never reset**

Once set, `plc_get_crash_signal()` will return the old signal number even after a new program is successfully loaded and runs normally. Consider resetting this to `0` when transitioning back to `PLC_STATE_RUNNING` (e.g., at the top of `plc_cycle_thread`).

core/src/plc_app/utils/watchdog.c
```diff
@@ -30,14 +30,23 @@ void *watchdog_thread(void *arg)
...
+            log_error("The loaded PLC program may contain an infinite loop. "
+                      "Upload a corrected program to recover.");
+
+            // Transition to ERROR state instead of killing the process.
+            // This keeps the runtime alive so the webserver can still
+            // communicate with it and upload a new program.
+            // Writing directly to plc_state is safe here because the PLC
+            // thread is unresponsive (stuck) and not contending the variable.
+            // The main loop checks plc_state each cycle and will exit.
+            plc_state = PLC_STATE_ERROR;
```
thiagoralves: **Concern: direct write to `plc_state` without holding `state_mutex`**

The comment says the PLC thread is stuck and not contending, but that's not guaranteed for all unresponsive cases. If the PLC thread is blocked in a slow I/O call or sleeping, it could wake up and race on `plc_state`. Every other state transition goes through the mutex -- this should too for correctness:

```c
pthread_mutex_lock(&state_mutex);
plc_state = PLC_STATE_ERROR;
pthread_mutex_unlock(&state_mutex);
```

core/src/plc_app/utils/watchdog.c
```diff
@@ -30,14 +30,23 @@ void *watchdog_thread(void *arg)
...
+            log_info("PLC State: ERROR");
+
+            // Reset heartbeat tracking for next run
+            last = 0;
+            atomic_store(&plc_heartbeat, 0);
+            continue;
         }
```
thiagoralves, line 49: **Issue: potential busy-loop error logging**

After setting ERROR state, the watchdog resets heartbeat and continues the loop. If the PLC thread is truly stuck in a tight infinite loop (no preemption point), it will never exit and never update the heartbeat. The watchdog will then re-detect "no heartbeat" every cycle and repeatedly log the same error messages.

Consider adding a guard to skip re-detection once ERROR state is set, e.g.:
```c
if (plc_state == PLC_STATE_ERROR) {
    // Already in error state, just wait
    continue;
}
```
at the top of the loop.

core/src/plc_app/plc_state_manager.h
```diff
@@ -32,4 +32,10 @@ bool plc_set_state(PLCState new_state);
  */
 void plc_state_manager_cleanup(void);
 
+/**
+ * @brief Get the signal number that caused the last PLC crash.
+ * @return The signal number (e.g. SIGFPE, SIGSEGV), or 0 if no crash occurred
```
thiagoralves, line 45: Nit: missing trailing newline at end of file.

core/src/plc_app/plc_main.c
```diff
@@ -36,6 +36,7 @@ void handle_sigint(int sig)
 int main(int argc, char *argv[])
 {
     bool print_debug = false;
+    bool safe_mode   = false;
```
thiagoralves, line 39: Nit: extra spaces for alignment (`safe_mode   = false`) -- the rest of the codebase uses single spaces.

webserver/runtimemanager.py
```diff
@@ -35,6 +39,8 @@ def __init__(self, runtime_path, plc_socket, log_socket, print_debug=False):
         self.runtime_socket = SyncUnixClient(plc_socket)
         self.monitor_thread = threading.Thread(target=self._monitor, daemon=True)
         self.running = False
+        self._crash_times: list[float] = []
```
thiagoralves, line 43: **Minor: thread safety of `_crash_times` and `_safe_mode`**

`_crash_times` is mutated in `_monitor` (daemon thread) and `reset_crash_tracking` (called from the compilation thread). These concurrent accesses are unprotected. This is largely safe under CPython due to the GIL, but it would be cleaner to protect with `self._lock` or add a comment documenting the GIL reliance.

webserver/runtimemanager.py
```diff
@@ -161,28 +167,56 @@ def is_runtime_alive(self):
...
+            if self.print_debug:
+                cmd.append("--print-debug")
+            if safe_mode:
+                cmd.append("--safe-mode")
+            self.process = subprocess.Popen(cmd)
+        except (OSError, subprocess.SubprocessError) as e:
+            logger.error("Failed to start PLC runtime process: %s", e)
+            self.process = None
+        time.sleep(1)  # Give time to start
+        self._safe_connect_runtime_socket()
```
thiagoralves, line 185: Nit: `_should_enter_safe_mode` both records the crash timestamp and checks the threshold, which is surprising for a method named `_should_*`. Consider renaming to something like `_record_crash_and_check_safe_mode` to make the side effect explicit, or split the append out to the call site.

PR #159, opened 2026-08-02
fix: Modbus/TCP: incorrect behavior when the connection is lost (issue #691)

JulioSergioFS:
## Problem

The OpenPLC Editor lets a Modbus TCP remote device pick, per I/O group, what happens to the
values when the device becomes unreachable: "Keep last value" or "Set to zero"
(openplc-editor#691). "Set to zero" never did anything anywhere in the stack.

On the runtime side there were two reasons for that:

- `ModbusIoPointConfig` had no field to receive the setting, so even a config carrying it would
  be ignored.
- On a failed read the polling loop logged the error, called `mark_disconnected()` and
  `continue`d. The point's last good value stayed in the IEC buffer for as long as the device
  was down — the program kept consuming stale data with no way to tell.

The editor half (emitting `error_handling` per I/O point) is a separate PR; this one accepts the
field and implements the behaviour.

## Changes

**`shared/plugin_config_decode/modbus_master_config_model.py`**
`ModbusIoPointConfig` gains `error_handling`, defaulting to `keep-last-value` — what every
config written before this field implies, and what the runtime did unconditionally. An
unrecognised value falls back to that default with a warning rather than raising: a config from
a newer editor must not take a whole device offline over one unknown string.

**`modbus_master/modbus_master_utils.py`**
Two pure helpers:

- `get_read_count_for_io_point` — how many registers/coils a read of the point asks for. This
  arithmetic was duplicated in both polling threads; it now lives in one place and sizes both
  the Modbus request and the zero-fill, so the two can't disagree.
- `get_zero_payload_for_io_point` — a zeroed payload shaped like the response it replaces:
  `False` for coils/discrete inputs, integer zeros for registers, respecting element width (a
  `%MD` point is two registers per element).

**`modbus_master/modbus_master_plugin.py`**
`queue_zero_fill_on_failure` pushes that payload onto the same `read_results_to_update` list a
successful read feeds. That is deliberate: the zero then travels through the same conversion and
the same batched, mutex-protected write as a real value, instead of introducing a second
buffer-writing path with its own locking to get wrong. Wired into all four communication-failure
paths (Modbus exception response, `response.isError()`, `ConnectionException`, and the catch-all)
in both read threads — `ModbusSlaveDevice` (one TCP device) and `ModbusBusHandler` (several
slaves on a shared RTU bus).

Points set to `keep-last-value` are untouched, so the default path behaves exactly as before.

Deliberately **not** covered: the `ValueError` raised for an invalid offset. That point is
misconfigured rather than disconnected, and zeroing would mask a configuration error.

## Testing

New `tests/pytest/modbus_master/test_error_handling_zero_fill.py` — 14 cases, no threads, no
sockets, no PLC: read counts per function code and IEC width, payload shape, and the queueing
honouring/ignoring the configured mode (including a point object with no `error_handling`
attribute at all, which must not start zeroing buffers).

`shared/plugin_config_decode/test_modbus_master_config_model.py` — 3 cases: the default, parsing
`set-to-zero`, and the fallback for an unknown value.

Baseline preserved: `tests/pytest/modbus_master` still reports the same 3 pre-existing failures
and 10 pre-existing errors (stale fixtures referencing methods that moved to
`modbus_master_memory` some refactors ago); passing tests went from 11 to 36.

### Manual verification

No hardware needed — a pymodbus slave on the host and the runtime in Docker:

1. Run a Modbus TCP slave with recognisable register values, bound to `0.0.0.0`.
2. Configure a remote device in the editor pointing at the docker bridge gateway
   (`172.17.0.1`), one read group (FC 3, offset 0, length 4) mapped to `%MW0`, cycle 500 ms,
   error handling **Set to zero**. Declare four located `WORD` variables so the debugger shows
   them.
3. Build + upload, start the PLC, confirm the values track the slave.
4. Kill the slave. The values must go to 0 within one cycle and stay there; bringing the slave
   back restores them.

Switching the group to "Keep last value" and repeating must show the values frozen instead —
that is what proves the field is being honoured rather than the plugin zeroing unconditionally.

## Known limitation

Only the points polled in the cycle where the failure surfaces get zeroed. With two groups on
the same device at different cycle times (say 500 ms and 10 s), the fast group is zeroed on the
failing cycle, then `ensure_connection` blocks in its reconnect loop and no further reads are
attempted — so the slow group keeps its stale value for the rest of the outage. Both groups
share the same configured behaviour, so this is visibly inconsistent.

The robust fix is to zero on the *transition* into the disconnected state — sweeping every
`set-to-zero` read point of the device — rather than per failed read. That needs the batch-flush
block (currently inline in both threads) extracted into a method, and is left out of this PR to
keep the tested change small. Happy to fold it in here if reviewers prefer.

Two smaller notes:

- On a shared RTU bus a single slave failing marks the whole bus disconnected, so points
  belonging to healthy slaves polled in the same cycle are also zeroed when configured that way.
  Their reads did genuinely fail, so this is defensible, but it is a consequence worth knowing.
- The implementation assumes the failure always surfaces through a read, which holds for the
  synchronous pymodbus client: after the peer closes the socket, `client.connected` stays `True`
  until a read is attempted, and the read raises `ConnectionException` (verified against
  pymodbus 3.11.2, the pinned version). An async client that detects EOF in the background would
  reach the reconnect path without any read failing, and the zero-fill would not run.

Fixes openplc-editor#691 (together with the editor and web PRs — without them the field never
reaches this code)


marconetsf (approved):
Image

Values going to zero when connection is lost. LGTM!

PR #161, opened 2026-08-04
fix(runtime): sync located globals from STruC++'s locatedGlobals[]

thiagoralves:
## The bug

`image_tables_bind_located_vars()` derived the config-scope located set **positionally**: it assumed STruC++ emitted program-local entries before `CONFIGURATION VAR_GLOBAL ... AT` entries, and took the tail not covered by any program's `located_range()` as the globals.

STruC++ emits config globals **first**, so the program-local block is always the tail. `covered_end` therefore reached `locatedVarsCount` as soon as **any** POU declared a located variable, `g_config_located_count` collapsed to `0`, and `image_tables_copy_config_globals_in/out` became no-ops — every located configuration global silently stopped being synced. `%MX` coils written over Modbus never reached the program, and `%IX`/`%QX`/`%MW` globals stopped being read and committed.

Reported on the forum as *"%MX locations now invalid - Runtime v4"*: the reporter's `%MX` writes worked until he added one unrelated located variable to a POU, and downgrading to **v4.1.3** — which predates the threaded process-image model — fixed it.

**Affected versions:** v4.1.4–v4.1.7 never copied located config globals at all; v4.1.8–v4.1.9 worked only while zero POUs declared a located variable. Nothing was logged in either case.

## Why the runtime could not fix this alone

The scope is unrecoverable here: globals are file-scope `GlobalVar<V>` singletons that appear nowhere in the object graph reachable through `ConfigurationInstance → Resource → Task → Program`. STruC++ now states it instead, emitting `locatedGlobals[]` with the canonical storage pointer of every located `VAR_GLOBAL` — the same `raw_ptr()` value it writes into `locatedVars[].pointer`.

## What changed

Replace the inferred slice with an **explicit index list built once at program load**, by joining `locatedVars[].pointer` against `locatedGlobals[]` on **pointer identity**. No ordering rule, no layout assumption, and no false positives since distinct objects have distinct addresses.

The join lives in `located_globals.c` behind a pointer-accessor callback, so it carries no STruC++ layout knowledge and is unit-testable.

Two validations guard the result:
- **Join totality** — every `locatedGlobals[]` entry must match some located variable, otherwise the generated arrays disagree.
- The program-range coverage bitmap is retained purely as an **independent cross-check witness**: it can flag a config global a task also claims, but it never decides a scope.

## Backwards compatible in both directions

STruC++ emits the `extern "C"` accessors beside its own array, so nothing here references a symbol an older program lacks — important because `scripts/compile.sh` rebuilds the shim against each upload's generated header, where a shim-side reference would be a **compile** error.

The accessors are resolved with `required=false` (same pattern as `strucpp_debug_locate`). An older program resolves NULL, in which case located configuration globals are not synced and a warning naming the fix is logged once at load. The program still runs and POU-local located I/O is unaffected. An older runtime never looks the symbols up and is bit-for-bit unaffected.

## Verification

Installed on a Raspberry Pi via the full `sudo ./install.sh`, driven entirely through the REST API (`/api/login` → `/api/upload-file?clean=1` → `/api/compilation-status` → `/api/start-plc`) exactly as the editor does. Test program: two located config globals (`%MX0.0`, `%MD0`) plus one POU-local located var (`%QX0.0`) — the shape that triggers the bug.

**Fix active:**
```
[image_tables] 3 located var(s) via copy-in/out (1 program-local, 2 config-scope shared globals)

initial       %MD0 = 9876     %MX0.0 coil = 0
coil 8192 ON  %MD0 = 1234     %MX0.0 coil = 1
coil 8192 OFF %MD0 = 9876
```
Both copy-in and copy-out confirmed over real Modbus TCP.

**Older program (no `locatedGlobals[]`):**
```
[WARNING] [image_tables] this program does not export locatedGlobals[] (built by an
older editor/STruC++) — located CONFIGURATION VAR_GLOBALs will NOT be synced;
re-export the project from a current OpenPLC Editor
[image_tables] 3 located var(s) via copy-in/out (3 program-local, 0 config-scope shared globals)
```
It compiled, started, and kept running — the program-local `%QX0.0` heartbeat toggled while `%MD0` stayed inert.

`tests/test_located_globals.c` covers order-independence, the multi-program shape, the degradation paths, inconsistency detection and unbound descriptors.

## Note for release

The fix activates only for programs **re-exported** from an editor shipping STruC++ v0.6.2 — the generated code carries `locatedGlobals[]`. Updating the runtime alone leaves the warning above rather than the fix, by design.

---

### Related PRs — merge together

This fix spans four repos. All target `development`.

| Repo | PR | Contents |
|---|---|---|
| STruCpp | Autonomy-Logic/STruCpp#202 | emits `locatedGlobals[]` + accessors; tightens located-address validation |
| openplc-runtime | Autonomy-Logic/openplc-runtime#161 | consumes it via pointer-identity join |
| openplc-editor | Autonomy-Logic/openplc-editor#983 | pins STruC++ v0.6.2 |
| openplc-web | Autonomy-Logic/openplc-web#643 | pins STruC++ v0.6.2 (shared parity) |

**Ordering:** the runtime degrades gracefully against an older STruC++ (warns, skips located globals, program still runs), and an older runtime ignores the new symbols entirely — so merge order does not matter. But the fix only becomes active for a project **re-exported** from an editor shipping v0.6.2, so all four need to land before it reaches users.

### DIFF
--- a/core/src/CMakeLists.txt
+++ b/core/src/CMakeLists.txt
@@ -40,6 +40,7 @@ add_executable(plc_main
     ${CMAKE_SOURCE_DIR}/core/src/plc_app/utils/utils.c
     ${CMAKE_SOURCE_DIR}/core/src/plc_app/utils/watchdog.c
     ${CMAKE_SOURCE_DIR}/core/src/plc_app/image_tables.cpp
+    ${CMAKE_SOURCE_DIR}/core/src/plc_app/located_globals.c
     ${CMAKE_SOURCE_DIR}/core/src/plc_app/journal_buffer.c
     ${CMAKE_SOURCE_DIR}/core/src/plc_app/debug_write_journal.cpp
     ${CMAKE_SOURCE_DIR}/core/src/plc_app/plc_io_cycle.cpp
--- a/core/src/plc_app/image_tables.cpp
+++ b/core/src/plc_app/image_tables.cpp
@@ -13,6 +13,7 @@
 
 extern "C" {
 #include "include/iec_python.h"
+#include "located_globals.h"
 }
 
 // Layout-compatible mirror of the strucpp ABI. The runtime executable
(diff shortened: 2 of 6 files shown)


JoaoGSP (approved):
LGTM

PR #162, opened 2026-08-05
feat(runtime): run/stop state control with a hardware mode switch

thiagoralves:
Run/stop state control for Runtime v4, plus two bugs found on hardware while testing it.

## 1. A hardware mode switch can gate run/stop (`a0289b2`)

`plc_switch` stores a position that any platform's VPP plugin pushes through `plc_set_switch_position()`. The runtime refuses a start while it reads STOP — hardware is authoritative no matter who asks, including a buggy plugin. Defaults to RUN, so a runtime with no switch-aware plugin behaves exactly as before. Plugins gained `request_plc_start`, `request_plc_stop`, `set_switch_position` and `get_plc_state`; the last maps the internal state onto the 0/1/2 values baremetal reports via FC 0x49, so vendor code can share one status-LED mapping across target types.

## 2. Stop now actually reinitialises the program (`3d160f5`)

A CTU that had counted to 100 came back still at 100 after stop→start, and the program appeared to run while producing nothing.

`dlclose()` was called and logged "PLC program unloaded successfully", but glibc will not unmap a DSO that defines `STB_GNU_UNIQUE` symbols — and strucpp's headers emit them for C++17 `inline` namespace-scope variables (`strucpp::GLOBALVAR`, `GLOBALBOOL`, `debug::type_ops`, `__CURRENT_TIME_NS`). The mapping stayed resident with every static intact, and because `find_libplc_file` hands `dlopen` the same path, glibc matched the still-loaded object by name and handed it straight back.

Fixed by compiling the program with `-fno-gnu-unique`. Verified safe: only the program `.so` defines those symbols (every plugin DSO and `plc_main` itself: zero), and the runtime deliberately crosses the DSO boundary through `ext_strucpp_*` pointers rather than symbol identity. The flag leaves them as `WEAK` at identical addresses, so one-instance-per-DSO is still enforced by the linker.

Measured on an SLM-RP4: with the flag the `.so` leaves `/proc/<pid>/maps` on every stop and re-maps on every start; without it, one build's image stays pinned for the life of the process.

## 3. The state itself is now the transition interlock (`b93d27b`)

Rapid run/stop flips could wedge the runtime permanently: STATUS stuck at TRANSITIONING, every other command answering `COMMAND:BUSY` (surfacing in the editor as "Could not verify MD5: Unexpected response format"), the editor showing stopped, and the program still scanning. Reproduced: 25 start/stop pairs 150 ms apart wedged at round 17.

Two sources of truth caused it. `plc_state` was written eagerly by whoever requested a change, with an `is_transitioning` atomic tracking overlap alongside it, and `plc_cycle_thread` then re-asserted RUNNING unconditionally once scheduled. A stop landing in the window before that thread first ran wrote STOPPED, joined the thread, and had RUNNING put back underneath it — after which no loop in that thread would ever exit, the join never returned, and the atomic was never cleared.

- `plc_claim_transition()` publishes `TRANSITIONING_TO_RUN` / `TRANSITIONING_TO_STOP` under the state lock and refuses any request while one is in flight; `plc_publish_final_state()` ends it. The atomic is gone.
- RUNNING is published immediately before the dispatcher releases the first scan. Both premature writes are gone, as is a mid-transition `PLC_STATE_INIT` write that made STATUS flicker to INIT and briefly hid that a transition was in flight.
- Dropped requests no longer lose the switch's intent: `plc_switch` records only that the switch *moved*, and the completion path compares where it came to rest against the state reached, correcting a mismatch. Movement-gated, so an editor Stop with the switch untouched is never reversed. Runtime-side, so no VPP needs changing.
- Watchdog backstop: TRANSITIONING past 60 s forces ERROR, since with the state as the interlock there is no flag anyone could clear.

The enum values are **appended**, never inserted — the first five are wire-visible through FC 0x46 and the plugin mapping.

## 4. Cleanup (`147b131`)

A plugin requesting a stop because someone moved the switch was logged at ERROR, so every ordinary flip painted the journal red. Now INFO. Two comments corrected, including `plc_main`'s claim that the auto-start gate sees a VPP-reported switch position — it does not, because that plugin initialises inside the start transition the gate is deciding about.

## Validation on hardware (SLM-RP4)

150 switch edges from rapid manual flipping produced:

| | |
|---|---|
| transitions started | 19 |
| landings (RUNNING/STOPPED) | 19 — all completed |
| reconciliations | 3 |
| watchdog stuck-forces | 0 |

150 edges collapsing to 19 transitions is the dropping working as intended. Reconciliation fired in both directions, including `came to rest in RUN but the PLC landed on STOPPED — correcting`, i.e. a dropped *start* recovered. Transitions land in 0.6–0.8 s (~4 s for a boot start, nearly all plugin bring-up).

## Notes for review

- `fb4cec0` and its revert `da408b9` are noise in the history (a plugin-ABI feature macro, added then removed in favour of component-level version control). Left as-is rather than rewriting a pushed branch.
- `development` is merged in (`ce35c40`); a post-merge rebuild on the device is running as of opening this PR.
- The paired editor/web change for "switch in STOP is a warning, not a failed upload" is openplc-editor#985 / openplc-web#645.

### DIFF
--- a/core/src/plc_app/plc_main.c
+++ b/core/src/plc_app/plc_main.c
@@ -194,14 +194,21 @@ int main(int argc, char *argv[])
         sleep(1);
     }
 
-    // Cleanup plugin driver system
+    log_info("Shutting down...");
+
+    // Stop the program BEFORE destroying the driver, not after.
+    //
+    // plc_state_manager_cleanup() tears the program down, and its teardown calls
+    // plugin_driver_stop(plugin_driver) -- so destroying the driver first left that
+    // call reading freed memory, and the cycle thread could still be running plugin
+    // cycle hooks through the same pointer while it wound down. The order here is
+    // the dependency order: no program, then no driver.
+    plc_state_manager_cleanup();
+
     if (plugin_driver)
     {
         plugin_driver_destroy(plugin_driver);
+        plugin_driver = NULL;
     }
-
-    // Cleanup
-    log_info("Shutting down...");
-    plc_state_manager_cleanup();
     return 0;
 }
--- a/core/src/plc_app/plc_main.c
+++ b/core/src/plc_app/plc_main.c
@@ -104,8 +104,9 @@ int main(int argc, char *argv[])
     wake_sa.sa_flags = 0;
     sigaction(SIGUSR1, &wake_sa, NULL);
 
-    // Make sure PLC starts in STOP state
-    plc_set_state(PLC_STATE_STOPPED);
+    // No need to force STOPPED here: plc_state is statically initialised to it,
+    // and plc_set_state() is now the body of a claimed transition rather than a
+    // setter -- calling it with nothing loaded would just log a failed unload.
 
     // Initialize watchdog
     if (watchdog_init() != 0)
@@ -156,7 +157,7 @@ int main(int argc, char *argv[])
     // faulty program that may have caused repeated crashes).
     //
     // Use plc_begin_transition() rather than plc_set_state() directly so the
-    // auto-start goes through the same is_transitioning CAS gate as any
+    // auto-start is arbitrated by plc_claim_transition() like any
     // socket-originated START command. This prevents a race where the socket
     // listener (already running above) accepts a START before the auto-start
     // finishes, causing two concurrent load_plc_program() calls — and two
--- a/core/src/plc_app/plc_main.c
+++ b/core/src/plc_app/plc_main.c
@@ -168,10 +168,16 @@ int main(int argc, char *argv[])
         log_info("Runtime started in SAFE MODE - PLC program will not be loaded");
         log_info("Upload a corrected program to recover");
     }
-    // Same gate as any other start. Plugin init ran above, so a VPP that owns a
-    // physical mode switch has already reported its position — a device powered
-    // up with the switch in STOP is gated here rather than starting and then
-    // being stopped a moment later.
+    // Same gate as any other start, but note what it can and cannot see. A VPP
+    // plugin that owns a physical mode switch is initialised as part of loading
+    // the program — inside the start transition below — so at this point the
+    // switch has usually NOT been reported yet and the gate reads the default
+    // (RUN). A device powered up with the switch in STOP therefore does start,
+    // and is then corrected: the plugin reports STOP during init, that request is
+    // dropped because a transition is in flight, and the switch-movement
+    // reconciliation stops the PLC as soon as the start lands. Safe, but the gate
+    // only bites here for a switch position already known at this point (e.g. one
+    // reported by a plugin the runtime loaded independently of the program).
     else if (!plc_switch_allows_run())
     {
         log_info("Hardware mode switch is in STOP - PLC left stopped");
--- a/core/src/plc_app/plc_main.c
+++ b/core/src/plc_app/plc_main.c
@@ -15,6 +15,7 @@
 #include "../drivers/plugin_driver.h"
 #include "image_tables.h"
 #include "plc_state_manager.h"
+#include "plc_switch.h"
 #include "plcapp_manager.h"
 #include "unix_socket.h"
 #include "utils/log.h"
@@ -165,6 +166,15 @@ int main(int argc, char *argv[])
     {
         log_info("Runtime started in SAFE MODE - PLC program will not be loaded");
         log_info("Upload a corrected program to recover");
+    }
+    // Same gate as any other start. Plugin init ran above, so a VPP that owns a
+    // physical mode switch has already reported its position — a device powered
+    // up with the switch in STOP is gated here rather than starting and then
+    // being stopped a moment later.
+    else if (!plc_switch_allows_run())
+    {
+        log_info("Hardware mode switch is in STOP - PLC left stopped");
+        log_info("Move the switch to RUN to start the PLC");
     }
     else if (!plc_begin_transition(PLC_STATE_RUNNING))
     {
--- a/core/src/plc_app/plc_main.c
+++ b/core/src/plc_app/plc_main.c
@@ -27,7 +27,10 @@ volatile sig_atomic_t keep_running = 1;
 plugin_driver_t *plugin_driver     = NULL;
 extern bool print_logs;
 
-void handle_sigint(int sig)
+/* Graceful shutdown for both signals that mean "stop": SIGINT from an
+ * interactive Ctrl-C, and SIGTERM from a supervisor. Drops out of the main loop
+ * so the program is stopped and the plugin driver torn down on the way out. */
+void handle_shutdown_signal(int sig)
 {
     (void)sig;
     keep_running = 0;
@@ -83,12 +86,22 @@ int main(int argc, char *argv[])
         return -1;
     }
 
-    // Handle SIGINT for graceful shutdown
+    // Handle SIGINT and SIGTERM for graceful shutdown.
+    //
+    // SIGTERM matters as much as SIGINT and was missing: RuntimeManager stops the
+    // runtime with process.terminate() (SIGTERM) and only escalates to
+    // process.kill() after a 5 s wait, and systemd's default KillSignal is also
+    // SIGTERM. With no handler installed the default disposition applied, so every
+    // supervisor-initiated stop killed the process outright -- the PLC program was
+    // never unloaded, plugins never stopped, the journal never flushed, and the
+    // program .so never dlclose'd. The escalation to SIGKILL is the backstop for a
+    // shutdown that hangs, not the normal path.
     struct sigaction sa;
-    sa.sa_handler = handle_sigint;
+    sa.sa_handler = handle_shutdown_signal;
     sigemptyset(&sa.sa_mask);
     sa.sa_flags = 0;
     sigaction(SIGINT, &sa, NULL);
+    sigaction(SIGTERM, &sa, NULL);
 
     // Install the process-wide SIGUSR1 wake handler exactly once. Task
     // threads (plc_state_manager.cpp) and EtherCAT bus threads
@@ -115,17 +128,20 @@ int main(int argc, char *argv[])
         return -1;
     }
 
-    // Start UNIX socket server
-    if (setup_unix_socket() != 0)
-    {
-        log_error("Failed to set up UNIX socket");
-        return -1;
-    }
-
     // Initialize plugin driver system BEFORE loading the PLC program.
     // plc_set_state(RUNNING) triggers load_plc_program() which uses the plugin
     // driver to update config and re-init plugins, and plc_cycle_thread() calls
     // plugin_driver_start() after image tables are populated.
+    //
+    // This block runs BEFORE the command socket exists, deliberately. A START
+    // accepted while it is still executing runs load_plc_program() ->
+    // plugin_driver_init() on the transition worker at the same time as this
+    // thread is inside plugin_driver_load_config()/plugin_driver_init(), and
+    // rebuilding a plugin slot dlcloses the .so -- so a plugin sleeping in its
+    // own init() on the other thread returns into an unmapped page. Observed as a
+    // SIGSEGV in the main thread at an address inside the plugin that had just
+    // been unloaded. Transition arbitration cannot help here: there is only one
+    // transition, racing driver setup rather than another transition.
     plugin_driver = plugin_driver_create();
     if (plugin_driver)
     {
@@ -146,11 +162,26 @@ int main(int argc, char *argv[])
         // This prevents a deadlock where the main thread holds the GIL forever
         // while sleeping, blocking other threads (like the unix socket thread)
         // from using Python when handling commands like START.
+        //
+        // Through the driver rather than PyEval_SaveThread() directly: the driver
+        // has to restore this exact thread state before Py_FinalizeEx() at
+        // shutdown, and the state was previously discarded here.
         if (Py_IsInitialized())
         {
-            PyEval_SaveThread();
+            plugin_driver_release_gil();
             log_info("[PLUGIN]: Released Python GIL");
         }
+    }
+
+    // Start the command socket only now that the plugin driver is fully built.
+    // Everything the socket can ask for -- START, STOP, PLUGIN_CMD, STATS --
+    // reaches into the driver, so serving commands before this point was serving
+    // them against a half-configured one. The webserver already tolerates the
+    // socket appearing a moment later: it connects, retries, and polls.
+    if (setup_unix_socket() != 0)
+    {
+        log_error("Failed to set up UNIX socket");
+        return -1;
     }
 
     // Start PLC (skip in safe mode to allow program upload without loading the
@@ -159,7 +190,7 @@ int main(int argc, char *argv[])
     // Use plc_begin_transition() rather than plc_set_state() directly so the
     // auto-start is arbitrated by plc_claim_transition() like any
     // socket-originated START command. This prevents a race where the socket
-    // listener (already running above) accepts a START before the auto-start
+    // listener (started just above) accepts a START before the auto-start
     // finishes, causing two concurrent load_plc_program() calls — and two
     // dispatcher threads. plc_begin_transition() also makes the start
     // asynchronous, which is fine: the main thread just sleeps below.
--- a/core/src/plc_app/plc_state_manager.cpp
+++ b/core/src/plc_app/plc_state_manager.cpp
@@ -442,10 +442,16 @@ void *plc_cycle_thread(void *arg)
 
     log_info("Starting main loop");
 
-    pthread_mutex_lock(&state_mutex);
-    plc_state = PLC_STATE_RUNNING;
-    pthread_mutex_unlock(&state_mutex);
-    log_info("PLC State: RUNNING");
+    /* NOT where RUNNING is published. The state stays TRANSITIONING_TO_RUN until
+     * the task threads exist and the dispatcher is about to release the first
+     * scan -- see the publish below the "Spawned N PLC task thread(s)" log. Two
+     * writes used to happen before this point (here, and in plc_set_state before
+     * this thread was even created), and both claimed RUNNING while nothing was
+     * scanning yet. The one here was also a deadlock: a stop landing in the
+     * window before this thread was first scheduled wrote STOPPED and joined us,
+     * and this line put RUNNING back -- after which no loop below would ever
+     * exit, so the join never returned and the runtime refused every command for
+     * the rest of the process's life. */
 
     clock_gettime(CLOCK_MONOTONIC, &timer_start);
 
@@ -708,6 +714,18 @@ void *plc_cycle_thread(void *arg)
     }
     log_info("Spawned %zu PLC task thread(s)", plc_task_count);
 
+    /* RUNNING, at last, and this is the earliest point it is true: the workers
+     * exist and the very next thing that happens is the dispatcher releasing the
+     * first scan. Nothing observes RUNNING too early as a result -- the workers
+     * are parked in sem_wait and are only ever posted from the loop below, and
+     * the loop itself needs RUNNING visible to run at all.
+     *
+     * This is also what ends the transition claimed by plc_claim_transition, so
+     * every path out of this thread from here on must land a final state: the
+     * crash recovery above publishes ERROR, and the stop path publishes STOPPED
+     * once teardown joins. */
+    plc_publish_final_state(PLC_STATE_RUNNING);
+
     /* ---------------------------------------------------------------------
      * GCD master-tick dispatcher.
      *
@@ -948,12 +966,14 @@ extern "C" int load_plc_program(PluginManager *pm)
 
     if (plugin_manager_load(pm))
     {
+        /* Progress, not a state change. This used to publish PLC_STATE_INIT,
+         * which is now actively wrong: it overwrites TRANSITIONING_TO_RUN, so the
+         * runtime would stop reporting a transition in flight and let a stop be
+         * claimed while the start was still landing -- the same hole the
+         * TRANSITIONING states exist to close. (It is also what made STATUS
+         * flicker to INIT mid-start.) INIT survives in the enum as a startup
+         * value; nothing writes it during a transition. */
         log_info("Loading PLC application");
-
-        pthread_mutex_lock(&state_mutex);
-        plc_state = PLC_STATE_INIT;
-        pthread_mutex_unlock(&state_mutex);
-        log_info("PLC State: INIT");
 
         if (plugin_driver)
         {
@@ -1047,14 +1067,18 @@ extern "C" int unload_plc_program(PluginManager *pm)
 {
     if (pm && pm == plc_program)
     {
-        PLCState prev_state = plc_get_state();
-
-        if (prev_state != PLC_STATE_ERROR)
+        /* The dispatcher and its workers leave their loops on anything that is
+         * not RUNNING, and the join below depends on that. A claimed stop has
+         * already published TRANSITIONING_TO_STOP, which is that signal. This
+         * also covers the shutdown path (plc_state_manager_cleanup), which tears
+         * down without claiming a transition first: publish the direction here so
+         * the loops still exit. ERROR is left alone -- it must survive teardown. */
+        pthread_mutex_lock(&state_mutex);
+        if (plc_state == PLC_STATE_RUNNING)
         {
-            pthread_mutex_lock(&state_mutex);
-            plc_state = PLC_STATE_STOPPED;
-            pthread_mutex_unlock(&state_mutex);
+            plc_state = PLC_STATE_TRANSITIONING_TO_STOP;
         }
+        pthread_mutex_unlock(&state_mutex);
 
         pthread_join(plc_thread, NULL);
 
@@ -1077,7 +1101,10 @@ extern "C" int unload_plc_program(PluginManager *pm)
         plc_program = NULL;
 
         log_info("PLC program unloaded successfully");
-        log_info("PLC State: STOPPED");
+
+        /* The teardown is done, so this is the moment STOPPED becomes true.
+         * plc_publish_final_state keeps ERROR if a task crashed on the way out. */
+        plc_publish_final_state(PLC_STATE_STOPPED);
         return 0;
     }
     else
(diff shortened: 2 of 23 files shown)


core/src/plc_app/utils/watchdog.c
```diff
@@ -12,16 +12,47 @@
...
+        // runtime in TRANSITIONING forever, refusing every command but PING and
+        // STATUS — the state is the interlock now, so there is no flag anyone
+        // could clear to recover. Every path is meant to land a final state;
+        // this is the backstop for the one that doesn't, turning a silent
+        // permanent wedge into a reported fault the webserver can act on.
+        if (current_state == PLC_STATE_TRANSITIONING_TO_RUN ||
+            current_state == PLC_STATE_TRANSITIONING_TO_STOP)
+        {
+            transitioning_ticks++;
+            if (transitioning_ticks * WATCHDOG_TICK_S > TRANSITION_STUCK_S)
```
marconetsf, line 53: **Forcing ERROR at 60 s re-opens the double-load window this PR closes.**

`TRANSITION_STUCK_S` is 60 s, but `unix_socket.c` budgets `LANDING_WAIT_MS = 90000` for the very same event — the two bounds disagree by 30 s, and the shorter one wins. Worse, `plc_force_error_state()` does not abort the in-flight transition; it only overwrites the state, which clears the interlock while `plc_set_state()` keeps running on the transition worker.

Concrete: a start whose plugin bring-up takes 65 s (EtherCAT scan, OPC-UA cert generation, a fieldbus probe on a cold bus) — well inside the 90 s the runtime itself considers normal. At t=60 the watchdog publishes ERROR. `handle_unix_socket_commands` now sees a non-transitioning, non-RUNNING state, so:

- a socket `START` is accepted → `plc_claim_transition(RUNNING)` succeeds from ERROR → a **second concurrent `load_plc_program()`** on the same `plc_program`, with `pthread_create` overwriting the still-live `plc_thread` handle;
- a plugin `request_plc_stop` is accepted → `unload_plc_program()` runs `pthread_join(plc_thread)` on a handle the first start has not yet assigned, and `plugin_manager_destroy()` on the manager the first start is still using.

The same applies to a slow stop (joining task threads with a long scan body). Suggest deriving both bounds from one constant with the watchdog strictly the looser of the two, e.g. `TRANSITION_STUCK_S >= LANDING_WAIT_MS/1000`.


---

**Relevance: HIGH — blocking.** Re-verified against the code. The constant ordering is inconsistent on its face (`TRANSITION_STUCK_S` 60 s vs `transition_worker`'s own `LANDING_WAIT_MS` 90 s), and `plc_force_error_state()` (`plc_state_manager.cpp:1271`) does nothing but write the state.

The realistic trigger is not a slow start: it is a **stop wedged in `pthread_join`** (`unload_plc_program:1083`) because a worker is in an infinite loop — precisely the condition the watchdog exists to catch. Sixty seconds later the state is ERROR, `plc_claim_transition(RUNNING)` accepts again (ERROR is neither transitioning nor RUNNING), and the next START runs `load_plc_program()` concurrently with the unload still joining.

core/src/plc_app/plc_state_manager.cpp
```diff
@@ -1047,14 +1067,18 @@ extern "C" int unload_plc_program(PluginManager *pm)
...
-
-        if (prev_state != PLC_STATE_ERROR)
+        /* The dispatcher and its workers leave their loops on anything that is
+         * not RUNNING, and the join below depends on that. A claimed stop has
+         * already published TRANSITIONING_TO_STOP, which is that signal. This
+         * also covers the shutdown path (plc_state_manager_cleanup), which tears
+         * down without claiming a transition first: publish the direction here so
+         * the loops still exit. ERROR is left alone -- it must survive teardown. */
+        pthread_mutex_lock(&state_mutex);
+        if (plc_state == PLC_STATE_RUNNING)
```
marconetsf: **Shutdown during a start now hangs forever.**

The comment says this guard "also covers the shutdown path (`plc_state_manager_cleanup`), which tears down without claiming a transition first" — but it only rewrites `RUNNING`. `plc_state_manager_cleanup()` is called from `plc_main.c:205` on SIGTERM, and if the boot auto-start (or any start) is still in flight the state is `TRANSITIONING_TO_RUN`, not `RUNNING`, so nothing is published here.

What follows: `pthread_join(plc_thread)` is entered while `plc_cycle_thread` is still bringing up; that thread then reaches `plc_publish_final_state(PLC_STATE_RUNNING)` (line 727) and enters `while (plc_get_state() == PLC_STATE_RUNNING)` — which now never becomes false, because the only writer that would have broken it already ran. The join never returns and `systemctl stop` / SIGTERM hangs until it is SIGKILLed, with the program still scanning.

This is exactly the resurrection the PR fixes for the claimed-stop path, still live on the unclaimed shutdown path. Two changes make it safe: widen the guard to any non-ERROR state, and make the RUNNING publish at line 727 conditional (see the separate comment there).

core/src/plc_app/plc_state_manager.cpp
```diff
@@ -708,6 +714,18 @@ void *plc_cycle_thread(void *arg)
...
+     * exist and the very next thing that happens is the dispatcher releasing the
+     * first scan. Nothing observes RUNNING too early as a result -- the workers
+     * are parked in sem_wait and are only ever posted from the loop below, and
+     * the loop itself needs RUNNING visible to run at all.
+     *
+     * This is also what ends the transition claimed by plc_claim_transition, so
+     * every path out of this thread from here on must land a final state: the
+     * crash recovery above publishes ERROR, and the stop path publishes STOPPED
+     * once teardown joins. */
+    plc_publish_final_state(PLC_STATE_RUNNING);
```
marconetsf: **`plc_publish_final_state(RUNNING)` is unconditional, so it can resurrect RUNNING over a state someone else already landed.**

The stickiness rule inside `plc_publish_final_state` only protects ERROR *against STOPPED*; RUNNING overwrites everything. Two reachable cases:

1. The watchdog forced ERROR because the transition exceeded `TRANSITION_STUCK_S` (60 s) — this line silently erases it and the runtime reports RUNNING after logging a hard fault.
2. `plc_state_manager_cleanup()` on shutdown left the state at `TRANSITIONING_TO_RUN` and is blocked in `pthread_join(plc_thread)` — this line publishes RUNNING and the dispatcher loop below never exits (see the comment on `unload_plc_program`).

Since the comment above states this line exists to *end the transition claimed by `plc_claim_transition`*, it should only fire when that transition is still the current one:

```suggestion
    if (plc_get_state() == PLC_STATE_TRANSITIONING_TO_RUN)
        plc_publish_final_state(PLC_STATE_RUNNING);
```

(with the caveat that the read+publish is not atomic; a `plc_publish_running_if_claimed()` helper doing both under `state_mutex` would be tighter.)

core/src/plc_app/unix_socket.c
```diff
@@ -43,56 +47,94 @@ static void *transition_worker(void *arg)
...
     }
 
     PLCState *arg = malloc(sizeof(PLCState));
     if (!arg)
     {
         log_error("Failed to allocate transition argument");
-        atomic_store(&is_transitioning, 0);
+        // Claimed but cannot run it: land a final state, or the runtime would sit
+        // in TRANSITIONING forever refusing every command.
+        plc_publish_final_state(PLC_STATE_STOPPED);
```
marconetsf: **Publishing STOPPED here reports a stop that never happened.**

Both failure paths (`malloc` at 131, `pthread_create` at 143) publish STOPPED unconditionally, but the claim has already published `TRANSITIONING_TO_STOP` — which is the signal the dispatcher and task loops use to leave `while (plc_get_state() == RUNNING)`. So when the target was STOPPED and the PLC was RUNNING:

- the scan threads exit and `plc_cycle_thread` returns,
- `unload_plc_program()` never runs, so `journal_cleanup()`, `plugin_driver_stop()`, `plugin_manager_destroy()`/`dlclose` are all skipped, `plc_program` stays non-NULL and `plc_thread` is never joined,
- STATUS reports `STOPPED`.

The next START then calls `load_plc_program(plc_program)` on a driver that was never stopped (`plugin_driver_init` re-entered on live plugin state) and overwrites `plc_thread` with a fresh `pthread_create`, leaking the unjoined one. `pthread_create` failing under thread/FD exhaustion is the realistic trigger.

Safer: on failure, restore the pre-claim state (or complete the transition synchronously on this thread) rather than asserting a landing that did not occur. Note also that for a failed *start* from EMPTY this publishes STOPPED, misreporting "no program" as "stopped".

core/src/plc_app/unix_socket.c
```diff
@@ -43,56 +47,94 @@ static void *transition_worker(void *arg)
...
+
+        // Only reconcile from a clean landing. ERROR and EMPTY are not states to
+        // "correct" — restarting a faulted or programless PLC because a switch
+        // moved would fight the fault rather than report it.
+        if ((landed == PLC_STATE_RUNNING || landed == PLC_STATE_STOPPED) && landed != wanted)
+        {
+            log_warn("Mode switch came to rest in %s but the PLC landed on %s — correcting",
+                     wanted == PLC_STATE_RUNNING ? "RUN" : "STOP",
+                     landed == PLC_STATE_RUNNING ? "RUNNING" : "STOPPED");
+            plc_begin_transition(wanted);
```
marconetsf: **The movement record is consumed before the correction is known to have started, and the return value is dropped.**

`plc_switch_take_movement()` at line 87 clears the record, then this call can still fail: `plc_claim_transition` refuses if another transition was claimed in the gap between the landing and here (the state is already final, so the socket thread is free to accept a START/STOP), and `plc_begin_transition` also returns false on the alloc/thread-create paths above. When it does, the switch's intent is gone — no retry, no record left to reconcile on the next transition, and not even a log line — leaving the PLC in the state the switch disagrees with, which is the exact failure mode this mechanism exists to prevent.

Suggest re-arming the record (or at minimum logging at warn) when the correcting transition is refused:

```suggestion
            if (!plc_begin_transition(wanted))
                log_warn("Mode switch reconciliation to %s was refused — switch and PLC still disagree",
                         wanted == PLC_STATE_RUNNING ? "RUN" : "STOP");
```

core/src/plc_app/unix_socket.c
```diff
@@ -212,10 +244,27 @@ void handle_unix_socket_commands(const char *command, char *response, size_t res
             strncpy(response, "STOP:ERROR\n", response_size);
         }
     }
+    else if (strcmp(command, "SWITCH") == 0)
```
marconetsf, line 308: **`SWITCH` answers `COMMAND:BUSY` during a transition — precisely when the editor is polling.**

The gate at the top of this function only allows `PING` and `STATUS` while `plc_state_is_transitioning()`, so `SWITCH` falls into `COMMAND:BUSY`. On the webserver side `parse_switch_position()` then returns `None`, so `handle_status` silently drops `switchPosition` from every status response for the whole duration of a start/stop, and `GET /switch` answers `"unknown"`. An editor that keys "is a start allowed" off this field loses it exactly while it is polling through the transition it just triggered.

`SWITCH` is a pure atomic read of `plc_switch` with no coupling to `plc_state`, so it is safe to answer mid-transition — it belongs in the allowlist next to `PING` and `STATUS`.

core/src/plc_app/plc_main.c
```diff
@@ -166,6 +168,21 @@ int main(int argc, char *argv[])
...
+    // plugin that owns a physical mode switch is initialised as part of loading
+    // the program — inside the start transition below — so at this point the
+    // switch has usually NOT been reported yet and the gate reads the default
+    // (RUN). A device powered up with the switch in STOP therefore does start,
+    // and is then corrected: the plugin reports STOP during init, that request is
+    // dropped because a transition is in flight, and the switch-movement
+    // reconciliation stops the PLC as soon as the start lands. Safe, but the gate
+    // only bites here for a switch position already known at this point (e.g. one
+    // reported by a plugin the runtime loaded independently of the program).
+    else if (!plc_switch_allows_run())
```
marconetsf, line 212: **A device that boots with the switch in STOP still drives its outputs for several seconds.**

The comment is honest that this gate reads the RUN default because the VPP plugin initialises inside the start transition — but the consequence is that the PLC actually executes scans and writes physical outputs from the moment `plc_cycle_thread` publishes RUNNING until the movement reconciliation lands the corrective stop: the PR measures ~4 s for a boot start plus 0.6-0.8 s for the stop. On a panel where the mode switch is the operator's interlock, "power up with the switch in STOP" is the one case that must never actuate anything, and it is the case this gate cannot catch.

The position *is* known before the first scan: `plugin_driver_init()` runs inside `load_plc_program()`, i.e. before `pthread_create(&plc_thread, ...)`. Re-checking `plc_switch_allows_run()` in `plc_cycle_thread` immediately before `plc_publish_final_state(PLC_STATE_RUNNING)` and aborting the start there (publish STOPPED, tear the workers back down) would close the window entirely instead of relying on after-the-fact correction.

core/src/plc_app/plc_state_manager.h
```diff
@@ -136,6 +157,32 @@ PLCState plc_get_state(void);
  */
 bool plc_set_state(PLCState new_state);
```
marconetsf, line 171: **Stale doc on `plc_set_state` — it is now a footgun for the next caller.**

The function is still exported and its doc still reads like a setter ("Set the PLC state... returns true if the state was changed"), but after this PR it must only ever be called *after* `plc_claim_transition()` has published a TRANSITIONING state, and it now publishes a final state on some paths. A caller following this doc (say a future plugin hook calling `plc_set_state(PLC_STATE_STOPPED)` directly) would tear the program down with no claim, no `TRANSITIONING_TO_STOP` for the scan loops on the RUNNING path, and would then publish STOPPED — corrupting the very interlock the PR introduces. Worth restating the precondition here, and ideally dropping it from the public header now that `plc_begin_transition` is the only intended entry point.

core/src/plc_app/utils/watchdog.c
thiagoralves: Partially addressed in \`eea93f1\`, and deliberately so — the constant inversion is gone, the abandonment is not.

The two bounds now come from one place: \`PLC_TRANSITION_STUCK_TIMEOUT_MS = PLC_TRANSITION_LANDING_TIMEOUT_MS + 30s\` in \`plc_state_manager.h\`, consumed by both \`unix_socket.c\` and \`watchdog.c\`. The watchdog can no longer fire while the runtime itself still considers the transition to be progressing, which removes the realistic trigger for the concurrent-load window you described.

What is **not** fixed is the substance of your second paragraph: \`plc_force_error_state()\` still only writes the state, so forcing ERROR releases the interlock while the worker that failed to land keeps running, and a START accepted from ERROR would begin a second transition over the top of it. That is recorded as a \`KNOWN LIMITATION\` comment on the function rather than papered over by the constant change.

Closing it properly needs the transition owner to be able to abort its own work, which is not something this function can do. Two findings from investigating that, since they change the shape of the fix:

- A runaway worker **can** be forced out, and the machinery already exists: \`plc_crash_handler\` + \`ctx->crash_jmp\` + the \`sigsetjmp\` recovery block already yank a worker out of arbitrary generated C++ and make it exit cleanly (releasing \`image_tables_mutex\` via \`holding_mutex\`, clearing \`alive\`, returning so the join reaps it). A forced abort is that same path on demand, via a dedicated signal — not \`SIGUSR1\`, which the EtherCAT threads depend on staying a no-op.
- The signaller has to outrank the target, and today it may not: \`plc_task_thread\` clamps a worker to 99 (line 199) while the dispatcher also runs at 99 under a comment claiming it is "above every worker". Equal priority under SCHED_FIFO means a spinning worker is never preempted. \`eea93f1\` clamps workers to 98 and reserves 99.

Tracking the abort ladder plus the single-owner refactor separately.

core/src/plc_app/plc_state_manager.cpp
thiagoralves: Fixed in \`eea93f1\`, and you were right that it needs both halves.

The guard is back to \`!= PLC_STATE_ERROR\` (which is what it was before this PR), and line 727 now goes through \`plc_publish_running_if_claimed()\` — which checks \`TRANSITIONING_TO_RUN\` and writes RUNNING in one critical section, and on refusal sends the cycle thread into its existing teardown block instead of the dispatcher loop. Widening the guard alone does leave the hang in place, exactly as you said.

Investigating this turned up two more defects on the same path, both pre-existing on \`development\`, fixed in \`d7f30ad\`:

- **\`plc_state_manager_cleanup()\` tore down on top of an in-flight transition.** During plugin bring-up \`load_plc_program\` has not assigned \`plc_thread\` yet, so \`pthread_join\` ran on a handle that was never set. It now waits for the transition to land first (bounded by the landing timeout), which turns shutdown-during-a-start into an ordinary stop.
- **SIGTERM had no handler at all.** So the premise of the comment is off for SIGTERM specifically: \`RuntimeManager\` stops the runtime with \`process.terminate()\` and systemd defaults to SIGTERM, and with no handler installed the default disposition applied — the process was killed outright rather than hanging, with the program never unloaded and the \`.so\` never \`dlclose\`d. The graceful path was SIGINT only. Both are handled now, and \`RuntimeManager\` waits 15 s instead of 5 so SIGKILL does not preempt the cleanup that SIGTERM now performs.

Verified in \`tests/lifecycle\` (\`c4d88c5\`): SIGINT and SIGTERM aimed into a start transition. Before the fixes the process dies on signal 11 and 15 respectively; after, both exit rc=0 in ~2.7 s with \`Shutdown waited ... for the state change in flight to land\` and \`PLC program unloaded successfully\` in the journal.

core/src/plc_app/plc_state_manager.cpp
thiagoralves: Fixed in \`eea93f1\` — implemented as the helper you suggested in the caveat, since the read-then-publish really is not atomic.

\`plc_publish_running_if_claimed()\` does the \`TRANSITIONING_TO_RUN\` check and the write in one \`state_mutex\` critical section, and returns false when the transition is no longer ours. The cycle thread then logs, reaps its workers through the extracted \`reap_task_threads()\`, and returns — it never enters the dispatcher loop, so there is nothing left to publish RUNNING over a landing someone else made.

Both cases you listed are covered: a watchdog-forced ERROR is no longer erased, and the shutdown path is no longer hung by it (with the widened guard in the sibling comment).

Verified in \`tests/lifecycle\`: the oracle is "RUNNING never left standing over a teardown" — RUNNING published with no matching unload in the journal fails the check.

core/src/plc_app/unix_socket.c
thiagoralves: Fixed in \`eea93f1\`, taking the second of your two suggestions.

Both failure paths now complete the transition synchronously on the calling thread (via an extracted \`run_transition()\`) instead of publishing a landing nothing reached. The first suggestion — restoring the pre-claim state — is not safe here: the claim has already published \`TRANSITIONING_TO_STOP\`, so by that point the dispatcher and workers may have left their loops, and republishing RUNNING would report a running PLC with no workers. Completing it inline blocks the caller for the duration, which on a thread-or-memory exhaustion path is the cheaper of the two costs; the socket is single-client so the editor simply waits.

Your note about a failed start from EMPTY reporting STOPPED is fixed by the same change — the transition now runs and publishes whatever it actually reaches.

This one is now **observed rather than argued**, which it could not be before: \`tests/lifecycle\` has an \`LD_PRELOAD\` shim that fails exactly one \`pthread_create\` (one-shot, self-disarming), and uses \`/proc/<pid>/maps\` as the oracle, since STATUS answers STOPPED either way and only the mapping tells the truth. Against the branch before the fix:

    [FAIL] STOP still lands a final state    STOP:ERROR -> STATUS:STOPPED
    [FAIL] the stop was real: program unmapped   still mapped -- STATUS lied about the stop
    [FAIL] teardown actually ran                 no unload in journal
    [PASS] the spawn failure was actually injected

After: \`STOP:OK\`, unmapped, unload logged. Exactly the phantom stop you described, including the part where the next start would re-enter \`plugin_driver_init\` on live plugin state and re-run a program whose statics were never reinitialised.

core/src/plc_app/unix_socket.c
thiagoralves: Fixed in \`eea93f1\`, going one step further than the suggested log line.

New \`plc_switch_note_movement()\` hands the record back when \`plc_begin_transition()\` is refused, so the next landing reconciles instead of the intent being lost, and there is a warn either way. Re-arming rather than only logging because a log line still leaves the PLC in the state the switch disagrees with. It cannot ping-pong: the retry compares position against state again, and a landing that agrees with the switch just consumes the record.

Worth being precise about what is verified. \`tests/lifecycle\` drives the full mechanism with a fake VPP plugin providing a file-driven mode switch — PLC running, flip to STOP (request accepted), flip back to RUN while the teardown is still in flight (request dropped), landing on STOPPED, reconciliation corrects it. 4/4 rounds reconcile with corrections logged. Getting there needed a deliberately slow \`stop_loop()\` in the fixture, since a stop is otherwise tens of milliseconds and there is nothing to aim at.

But that exercise passes on the unfixed branch too, because the corrective transition is not refused in it. The specific branch this fix adds only runs when the claim is refused by a request that slipped into the gap, and I could not force that deterministically without adding a hook to the runtime. So: mechanism verified end-to-end, the re-arm branch itself still argued from code. Listed as such in \`tests/lifecycle/README.md\` rather than left implicit.

core/src/plc_app/unix_socket.c
thiagoralves: Fixed in \`eea93f1\` — \`SWITCH\` is in the allowlist next to \`PING\` and \`STATUS\`, via an extracted \`format_switch_response()\` so the two call sites share one body. Your reasoning was exactly right: it is a plain atomic load of \`plc_switch\` with no coupling to \`plc_state\`, so there is nothing mid-change for it to expose.

Verified end to end through the webserver rather than just at the socket, since the symptom you described was on that side:

    before:  switchPosition present throughout -- 0 present, 6 missing
    after:   switchPosition present throughout -- 7 present, 0 missing

and the socket-level check in \`tests/lifecycle\` goes from \`0 answered, 123 BUSY\` to \`123 answered, 0 BUSY\`.

core/src/plc_app/plc_state_manager.h
thiagoralves: Fixed in \`eea93f1\`, keeping the declaration but making it say what it is.

The doc now states the precondition outright — \`plc_claim_transition()\` must already have published a TRANSITIONING state — spells out what a direct call would corrupt, and points callers at \`plc_begin_transition()\` as the only intended entry point. It also drops the "returns true if the state was changed" phrasing, which was the actively misleading part.

Not dropped from the header: \`plc_begin_transition()\` lives in \`unix_socket.c\` and is its only caller, so \`static\` is not available without moving the transition orchestration into \`plc_state_manager.cpp\`. That move is the right one and is the second phase of the single-owner refactor — the orchestration was in \`plc_state_manager.cpp\` originally and left in \`b54a592\` ("Robust plugin stop sequence with non-blocking socket commands") to make the socket non-blocking. The async requirement was real; siting the orchestrator in the transport was incidental. A standing executor thread in \`plc_state_manager.cpp\` satisfies both, and at that point these functions leave the public header entirely.

core/src/plc_app/plc_main.c
thiagoralves: Not fixed in this PR — deliberately, and I want to be explicit that the concern is valid rather than dismissed. A device powered up with the switch in STOP does execute scans and write physical outputs for the duration of the start (~4 s on an SLM-RP4) plus the corrective stop (0.6-0.8 s). On a panel where the switch is the operator interlock, that is the case that should never actuate.

Your proposed fix is right about the facts: `plugin_driver_init()` does run inside `load_plc_program()` before `pthread_create(&plc_thread, ...)`, so the position generally *is* known by the time the cycle thread is about to publish RUNNING, and re-checking there would close most of the window.

The reason it is not in this PR is that the abort half is not safely expressible yet. Publishing STOPPED from `plc_cycle_thread` and tearing the workers back down leaves `plc_program` non-NULL, the plugins initialised and the program `.so` still mapped — the same partial teardown as the phantom-stop finding on this PR, which also means the next start would re-run a program whose statics were never reinitialised, undoing the `-fno-gnu-unique` fix this PR is partly about. A correct veto has to route the abort through the real stop path, and no single component currently owns both directions: the socket thread orchestrates transitions, the cycle thread owns bring-up, and neither can perform the other's teardown.

That is the second phase of the single-owner refactor (a standing lifecycle executor in `plc_state_manager.cpp`, which is where the orchestration lived before `b54a592` moved it out to make the socket non-blocking). Once the executor owns both directions, the veto is a check at the point where it decides to publish RUNNING, and the abort is just "run the stop transition" — a few lines, and safe.

Two caveats for whoever picks it up, so the ticket is honest about its limits:

- It is best-effort, not a guarantee. It only helps for plugins that report the position synchronously inside `init()`; one that reports from its own thread or a GPIO interrupt after `init()` returns can still land after the check.
- The truly safe version does not start at all. Gating before `load_plc_program()` requires the position to be known before the program is loaded, which means a switch-aware plugin the runtime loads independently of the program — the case the gate comment already says it does bite for.

The current behaviour is documented at the gate in `plc_main.c` rather than left implicit, and the reconciliation does recover it. Filing the veto against the executor work.

thiagoralves:
## Review findings addressed — `eea93f1`, `d7f30ad`, `c4d88c5`

All eight inline threads answered and resolved. Seven fixed, one deferred with justification (finding 7, the boot-with-switch-in-STOP window), and finding 1 partially: the constant inversion is gone, the abandonment behind `plc_force_error_state()` is not, and that is now a `KNOWN LIMITATION` comment on the function rather than something the constant change pretends to close.

### `eea93f1` — the review findings

| Finding | Fix |
|---|---|
| 2 + 3 | `plc_publish_running_if_claimed()` checks `TRANSITIONING_TO_RUN` and writes RUNNING in one critical section; on refusal the cycle thread reaps its workers and exits. Paired with the `unload_plc_program` guard restored to `!= PLC_STATE_ERROR`. Both are needed — either alone leaves the hang. |
| 1 | `PLC_TRANSITION_STUCK_TIMEOUT_MS = PLC_TRANSITION_LANDING_TIMEOUT_MS + 30s`, one definition, two consumers. |
| 4 | Spawn-failure paths complete the transition inline instead of publishing a landing nothing reached. |
| 5 | `plc_switch_note_movement()` re-arms the record when a correction is refused. |
| 6 | `SWITCH` joins `PING`/`STATUS` in the mid-transition allowlist. |
| 8 | `plc_set_state()`'s doc states its precondition and points at `plc_begin_transition()`. |

Plus: the dead `extern PLCState plc_state` in `unix_socket.c` (the definition is `static`, so nothing could ever have linked to it), and worker SCHED_FIFO priority clamped to 98 so the dispatcher's 99 is actually above every worker rather than tying one.

### `d7f30ad` — graceful shutdown, four defects, all pre-existing on `development`

Found while building a test for finding 2: the runtime could not survive the path those fixes are about.

- **SIGTERM had no handler.** `RuntimeManager` stops the runtime with `process.terminate()`, and systemd defaults to SIGTERM, so every supervisor-initiated stop killed the process outright — program never unloaded, plugins never stopped, journal never flushed, `.so` never `dlclose`d.
- **`Py_FinalizeEx()` ran without the GIL**, segfaulting on every graceful shutdown of a runtime whose PLC had never started — including safe mode, i.e. the crash-recovery path. `main_tstate` was only ever assigned by `plugin_driver_start()`; `plc_main` called `PyEval_SaveThread()` and discarded the result. So a runtime that had run once exited cleanly and one that had not crashed in `PyImport_GetModule`.
- **The command socket accepted commands before the plugin driver was built.** A START in that window ran `plugin_driver_init` on the transition worker while the main thread was still in `plugin_driver_load_config`, and rebuilding a slot `dlclose`s the `.so` — a plugin sleeping in its own `init()` returned into an unmapped page. Seen as a SIGSEGV in the main thread inside the just-unloaded plugin.
- **`plc_state_manager_cleanup()` tore down on top of an in-flight transition**, joining a `plc_thread` that bring-up had not assigned yet. It now waits for the landing first.

`RuntimeManager` waits `RUNTIME_SHUTDOWN_TIMEOUT_S = 15` before SIGKILL, since 5 s would preempt the cleanup SIGTERM now performs (~4 s boot start on an SLM-RP4).

### `c4d88c5` — `tests/lifecycle`

Boots a real `plc_main` with a real compiled program, drives it over the command socket, and judges it by what an outside observer can see — socket replies, the journal, and `/proc/<pid>/maps`. That last one matters: STATUS answers STOPPED whether or not the stop unloaded anything.

**24 checks. 9 fail against the branch before these fixes; 0 after.** Two fixtures make the unreachable reachable: a fake VPP plugin with configurable sleeps in `init()`/`stop_loop()` (a container brings up in ~150 ms and stops in tens of ms — nothing to aim at) and a file-driven mode switch, plus an `LD_PRELOAD` shim that fails exactly one `pthread_create`.

The phantom stop is now observed rather than argued — before the fix: `STOP:ERROR`, `.so` **still mapped**, no unload in the journal, with the injection confirmed to have fired.

Still argued from code, and listed as such in the README rather than left implicit: the re-arm branch of reconciliation (needs a refusal that cannot be forced without a hook), the watchdog forcing ERROR (needs a transition outliving a 2-minute bound), and terminating a runaway IEC task (needs an abort ladder that does not exist yet).

### Not in this PR

- **Finding 7**, the boot-with-switch-in-STOP window — a veto inside `plc_cycle_thread` would leave the same partial teardown as finding 4. Needs one component owning both directions.
- **The single-owner refactor**, which is what removes this class rather than these instances. `plc_set_state()` in `plc_state_manager.cpp` *was* the single entry point; the orchestration left in `b54a592` to make the socket non-blocking. A standing lifecycle executor there satisfies both, and takes the remaining four protocol functions out of the public header.

### Verification

Linux build clean, zero warnings. `pytest`: 53 passed, with the same 3 pre-existing `modbus_master` mock failures present on `development`. Not yet re-run on hardware — the SLM-RP4 should confirm the shutdown timings, since every measurement above is from a container.

PR #165, opened 2026-08-10
fix(opcua): write REAL/LREAL values, not their bit pattern

JulioSergioFS:
Fixes the forum report "Error changing values via OPC UA" · RTOP-<n>

## Problem

Writing a floating-point value from an OPC-UA client stored garbage in the PLC.
A user wrote 42 to INT, REAL and LREAL variables: INT was correct, REAL read
back 1.109918e+09 and LREAL 4.6311077918204232e+18. Reading worked, and forcing
the value from the IDE debugger worked -- only client writes were wrong.

Those numbers are the IEEE-754 bit pattern of 42.0 read as an integer
(float32 -> 1109917696, float64 -> 4631107791820423168). The reporter's own IDE
screenshot confirms the PLC really holds `1109917696.000000` in a REAL.

## Root cause

The client write path is a single conversion hop:

    client 42.0 -> convert_value_for_plc('REAL', 42.0) -> debug_write_value(...)

`convert_value_for_plc` packed the float into its integer bit pattern
(`struct.unpack('I', struct.pack('f', value))`) and `debug_write_value` then
encoded that integer with `ctypes.c_float` -- storing 1109917696.0.

That packing was correct under the MatIEC-era debug protocol, which exchanged
variables as raw integers. When the plugin moved onto the STruC++ debug surface,
reads and writes became typed floats (`c_float` / `c_double`), but these two
helpers kept the old contract -- their comments still asserted "Float/Real
values are stored as integers in debug variables". This is why the reporter
notes the previous runtime version did not have the issue.

**Not endianness** (the initial hypothesis): byte-swapping 42.0f yields
1.44e-41 (10306 as an integer), not the reported values; INT writes were
unaffected; and the reporter reproduced it on two x86_64 hosts (Docker on
Ubuntu 24.04 and a mini-pc on Debian 13), which answers the platform question
raised in the thread.

## Changes

`core/src/drivers/plugins/python/opcua/opcua_utils.py` (+19 / -40):

- `convert_value_for_plc`: FLOAT/REAL/LREAL merge into one branch returning
  `float(value)`. This is the fix.
- `convert_value_for_opcua`: same merge, dropping the mirrored assumption
  (`isinstance(value, int)` decoded through `struct.unpack`). No observable
  behaviour change -- `debug_read_value` always returns a float and node seed
  values are `0.0` -- but leaving the read side decoding integers while the
  write side stops producing them would re-open this bug.
- Float exception fallback returns `0.0` instead of int `0`; the now-unused
  `struct` import is gone.

## Testing

`tests/pytest/plugins/opcua/test_type_conversions.py` asserted the buggy
contract (packing/unpacking bit patterns) and had no LREAL write coverage at
all. Those tests were rewritten, plus a regression pinned to the reported
values (42.0 in, 42.0 out, explicitly not 1109917696 / 4.6311077918204232e+18).

New `tests/pytest/plugins/opcua/test_write_path.py` exercises the full cycle --
`convert_value_for_plc` -> `debug_write_value` -> memory -> `debug_read_value`
-> `convert_value_for_opcua` -- against a fake `args` that keeps the bytes in a
dict like `plc_main` does. Nothing tested that pair before, which is how the
mismatch survived. It covers REAL/LREAL round-trips, the REAL float32 rounding,
an integer variant sent to a Float node, the exact bytes reaching memory, and a
non-regression class for integers and BOOL.

- 27 new tests pass; the two files together run 119 green.
- The new tests fail against the pre-fix code (20 failures when run against a
  copy with the original `opcua_utils.py`), so they pin the behaviour.
- Whole OPC-UA test folder: 48 failed / 170 passed before, 48 failed / 197
  passed after -- the same pre-existing failures, none new. Those 48 come from a
  stale fixture (`opcua_datatype_test.json` has no `format_version`, and the
  runtime requires >= 2) plus rate-limit tests in `test_user_manager.py`; they
  reproduce on a clean checkout and deserve their own ticket.
- `ruff` reports nothing on the touched code (the two F401s in `opcua_utils.py`
  are pre-existing, in the import fallback). The pre-existing files were
  deliberately not reformatted -- they are not black-clean at HEAD and
  reformatting would bury the fix in noise.

Manual verification: write 42.0 to a REAL and an LREAL node from an OPC-UA
client (with an explicit Float / Double variant) and confirm 42 both on
read-back and in the IDE debug view; write to an INT to confirm no regression.
The plugin is pure Python -- nothing to recompile -- but replacing the file in a
running container does not reload the already-imported module: restart the
container first.

## Related findings (separate tickets)

- TIME reads carry the same stale contract: `debug_read_value` returns int64
  nanoseconds while `convert_value_for_opcua("TIME", int)` assumes milliseconds
  and passes the value through -- a 1e6 factor error towards the client.
- Integer writes still clamp out-of-range values (40000 into an INT becomes
  32767) instead of the two's-complement wrap C/IEC perform. That is the
  mechanism behind openplc-editor#619; it changes live-path behaviour, so it
  needs its own ticket.
- `DOPE-251`, listed as related prior work, does not cover this path: no commit
  in either repo references it, and the endianness fixes that do exist are
  unrelated (EtherCAT SDO encoding; the Modbus MD5 sentinel in the editor).

### DIFF
--- a/core/src/drivers/plugins/python/opcua/opcua_utils.py
+++ b/core/src/drivers/plugins/python/opcua/opcua_utils.py
@@ -3,7 +3,6 @@
 import ctypes
 import os
 import sys
-import struct
 from typing import Any
 from asyncua import ua
 
@@ -140,24 +139,13 @@ def convert_value_for_opcua(datatype: str, value: Any) -> Any:
             clamped_value = max(0, min(18446744073709551615, int(value)))
             return ctypes.c_uint64(clamped_value).value
 
-        elif datatype.upper() in ["FLOAT", "REAL"]:
-            # Float/Real values are stored as integers in debug variables
-            # Convert back to float if it's an integer representation
-            if isinstance(value, int):
-                try:
-                    return struct.unpack('f', struct.pack('I', value))[0]
-                except struct.error:
-                    return float(value)
-            return float(value)
-
-        elif datatype.upper() == "LREAL":
-            # LREAL (64-bit float) values are stored as integers in debug variables
-            # Convert back to double if it's an integer representation
-            if isinstance(value, int):
-                try:
-                    return struct.unpack('d', struct.pack('Q', value))[0]
-                except struct.error:
-                    return float(value)
+        elif datatype.upper() in ["FLOAT", "REAL", "LREAL"]:
+            # debug_read_value casts the raw bytes to c_float / c_double, so
+            # what arrives here is the number itself, never its bit pattern.
+            # The MatIEC-era debug protocol did hand over raw integers, and the
+            # struct.unpack that decoded them outlived it -- on the STruC++
+            # debug surface it would turn an integer-valued REAL of 1 into
+            # 1.4e-45. Mirror of the write-side fix below.
             return float(value)
 
         elif datatype.upper() == "STRING":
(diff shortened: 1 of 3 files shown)


marconetsf (approved):
Everything running accordingly.

PR #169, opened 2026-08-19
feat(license): VPP license delivery and enforcement hardening for runtime-v4

marconetsf:
## Why

This is the **runtime half of licensing VPPs on runtime-v4 targets**. The editor half is open as Autonomy-Logic/openplc-editor#1023 (mirror: Autonomy-Logic/openplc-web#681); the closed license-core and the key model live in openplc-packages (ADR-0004, merged). Without this branch, a paid runtime-v4 VPP compiles and uploads but can never be activated: it runs its 2-hour demo forever.

## What (9 commits, in dependency order)

**License transport (0x48/0x49/0x4A) at the webserver level**
- `webserver/vpp_license_debug.py` answers the same Modbus PDUs the Arduino firmware speaks — byte-identical wire contract, so the editor has ONE licensing flow on every medium — resolved **ahead of the `is_connected` gate**, so a device can be activated while the PLC is stopped (the chicken-and-egg of licensing before the first run).
- 0x48 returns the hardware anchor (`/proc/device-tree/serial-number`), normalized EXACTLY as the closed core's C does (trailing NUL/LF/CR/space; >64 bytes refused, never truncated onto the wire).
- 0x49 validates the blob before touching the filesystem (size/magic/crc32, same status bytes as the bare-metal store) and writes **atomically** (tmp + fsync + `os.replace`) — a power cut mid-activation can no longer destroy the previous valid license. 0x4A validates on read too, so a cloned/torn 98-byte file reports CORRUPT instead of SUCCESS (which used to suppress the editor's one automatic repair path).
- Bundle delivery: `apply_vpp_plugin_conf` installs `conf/<plugin>.license` shipped inside an upload.
- Path containment for everything an upload can name (realpath + commonpath inside the runtime root; symlink-safe), and one shared license-path derivation that mirrors the plugin C exactly.

**Package signature verification before compile**
- A licensed VPP ships closed objects and a link-only Makefile that `compile.sh` runs as root — and nothing proved where those bytes came from: a 3-line `license_gate.c` stub in the upload rebuilt `license_gate.o` and ran FULL (measured, not assumed). Now: Ed25519 verification of the package's `signature.json` over the canonical payload before anything is installed; a verification **seal** whose tree digest `compile.sh` re-checks before running the uploaded Makefile; and `vpp_plugin_seal.c` re-checking the `.so`'s sha256 immediately before `dlopen`. Unsigned plain PLC programs are untouched. Scope: this closes the **licensing bypass**, not the pre-existing native-code-as-root surface (a privilege question, documented in the commit).

**Debug channel as a trust boundary**
- Every debug command re-runs the full JWT pipeline (expiry, revocation, user lookup) instead of trusting the connect-time check; the license FCs additionally require **admin** — a plain user can still debug variables but cannot read the anchor or overwrite a license.

**Cross-language contract tests**
- Anchor normalization pinned against the REAL C: the test compiles `license-core/src/license_platform.c` unmodified (through its own `LIC_LINUX_ANCHOR_PATH` seam) and compares byte-for-byte with the Python across an edge-case table. Rebase note: the canonical C moved from `rpi_plugin.c` into the closed core (ADR-0003) while this branch sat unmerged — the final commit follows it, and the repoint made the test stronger (whole TU, zero transcription). Skips on a runtime-only checkout (`OPENPLC_PACKAGES_DIR` overrides).
- The python/bash tree-digest pair (`vpp_package_signature.py` × `compile.sh`) pinned by executing the real bash functions, with mutation checks.
- The blob round-trip asserted against the signed golden blob from license-core — an artifact another implementation produced.

## Rebase

Rebased today onto `development` (v4.1.10 + run/stop + user-management), **clean — zero conflicts**; only the test repoint above was needed.

## Verification

- Full runtime pytest on the rebased branch: **161 passed** (modbus_master/slave and opcua dirs skipped locally — optional deps absent from the venv, untouched by this branch).
- VPP-focused suites: delivery, debug FCs, signature, tree digest, anchor cross-language, websocket auth — **92 passed**.
- Known limitation, documented in-code: the license FCs act on the FIRST plugin with a `config_path` in `vpp_plugins.conf` (the PDU carries no plugin id). A free VPP installed ahead of the licensed one would receive the blob — WARNed at runtime; install the licensable VPP alone or first.
- Real-device (Pi) bench run still pending — everything above is pinned by tests, not by hardware.

https://claude.ai/code/session_015uUH3ZL5ehreMUf2dtanWD

### DIFF
--- a/core/src/drivers/plugin_config.c
+++ b/core/src/drivers/plugin_config.c
@@ -21,7 +21,65 @@ static void remove_newline(char *str)
     }
 }
 
-int parse_plugin_config(const char *config_file, plugin_config_t *configs, int max_configs)
+/**
+ * Reject a plugin path that could point outside the runtime tree.
+ *
+ * The `path` field of a plugin config is handed straight to dlopen (VPP) or
+ * used as a Python module location, so a config the runtime did not write is
+ * arbitrary-code selection. vpp_plugins.conf IS such a config: it arrives
+ * verbatim inside the user's upload. The Python side contains it too
+ * (webserver/plcapp_management.py validate_vpp_plugins_conf); this check is
+ * here so containment does not depend on one language alone.
+ *
+ * @param require_contained 0 to only reject `..` traversal, 1 to also reject
+ *        absolute paths. Config files the runtime itself owns (plugins.conf)
+ *        pass 0: an operator with a hand-written absolute path there is not
+ *        the threat, and refusing it would break working installations. Only
+ *        the upload-supplied config is parsed with 1.
+ * @return 1 when the path is acceptable, 0 when it must be rejected.
+ */
+static int plugin_path_is_acceptable(const char *path, int require_contained)
+{
+    if (!path || path[0] == '\0')
+    {
+        return 0;
+    }
+
+    /* Any ".." component escapes, in every config, with no legitimate use. */
+    const char *cursor = path;
+    while (*cursor != '\0')
+    {
+        if (cursor[0] == '.' && cursor[1] == '.' &&
+            (cursor[2] == '\0' || cursor[2] == '/' || cursor[2] == '\\'))
+        {
+            /* Only a ".." that starts a component counts, so a file legally
+             * named "libfoo..so" is not rejected. */
+            if (cursor == path || cursor[-1] == '/' || cursor[-1] == '\\')
+            {
+                return 0;
+            }
+        }
+        cursor++;
+    }
+
+    if (require_contained)
+    {
+        if (path[0] == '/' || path[0] == '\\')
+        {
+            return 0;
+        }
+        /* Windows-style drive prefix ("C:\..."), reachable on the Cygwin build. */
+        if (path[1] == ':')
+        {
+            return 0;
+        }
+    }
+
+    return 1;
+}
+
+static int parse_plugin_config_internal(const char *config_file, plugin_config_t *configs,
+                                        int max_configs, int require_contained)
 {
     FILE *file = fopen(config_file, "r");
     if (!file)
@@ -56,6 +114,17 @@ int parse_plugin_config(const char *config_file, plugin_config_t *configs, int m
         strncpy(configs[config_count].path, token, sizeof(configs[config_count].path) - 1);
         configs[config_count].path[sizeof(configs[config_count].path) - 1] = '\0';
         remove_newline(configs[config_count].path);
+
+        /* Containment: drop the whole entry rather than load from a path that
+         * escapes the runtime tree. Skipping the entry (instead of aborting the
+         * parse) keeps one bad line from disabling every other plugin. */
+        if (!plugin_path_is_acceptable(configs[config_count].path, require_contained))
+        {
+            log_error("[PLUGIN] rejected plugin '%s' from %s: path '%s' is not contained in the "
+                      "runtime tree",
+                      configs[config_count].name, config_file, configs[config_count].path);
+            continue;
+        }
 
         // Parsing enabled
         token = strtok(NULL, ",");
@@ -109,4 +178,15 @@ int parse_plugin_config(const char *config_file, plugin_config_t *configs, int m
 
     fclose(file);
     return config_count;
+}
+
+int parse_plugin_config(const char *config_file, plugin_config_t *configs, int max_configs)
+{
+    return parse_plugin_config_internal(config_file, configs, max_configs, 0);
+}
+
+int parse_plugin_config_contained(const char *config_file, plugin_config_t *configs,
+                                  int max_configs)
+{
+    return parse_plugin_config_internal(config_file, configs, max_configs, 1);
 }
--- a/scripts/compile.sh
+++ b/scripts/compile.sh
@@ -139,9 +139,9 @@ VPP_OBJECT_SEAL="$VPP_OUTPUT_DIR/vpp_plugin.seal"
 # fails the build instead.
 sha256_hex() {
     if command -v sha256sum > /dev/null 2>&1; then
-        sha256sum "$1" | cut -d' ' -f1
+        sha256sum "$1" | awk '{ sub(/^\\/, "", $1); print $1; exit }'
     elif command -v shasum > /dev/null 2>&1; then
-        shasum -a 256 "$1" | cut -d' ' -f1
+        shasum -a 256 "$1" | awk '{ sub(/^\\/, "", $1); print $1; exit }'
     else
         return 1
     fi
@@ -153,14 +153,24 @@ sha256_hex() {
 # the value in the seal -- hence LC_ALL=C for byte-order sort, POSIX relative
 # paths, and exactly two spaces.
 vpp_tree_digest() {
-    local dir="$1" f inner
-    (
+    local dir="$1" listing
+    # '|| exit 1' used to exit the while's SUBSHELL (a pipeline stage); the
+    # function's status was cut's (0), so a hashing failure emitted a digest
+    # of a PARTIAL listing -- reported as "tree changed after it was
+    # verified", accusing an intact package of tampering when the real cause
+    # was a missing tool (review 2026-08-20, R4). pipefail makes the inner
+    # failure the pipeline's, and the caller sees it.
+    listing=$(
         cd "$dir" || exit 1
+        set -o pipefail
         find . -type f -print | sed 's|^\./||' | LC_ALL=C sort | while IFS= read -r f; do
-            inner=$(sha256_hex "$f") || exit 1
+            inner=$(sha256_hex "$f") || { echo "[ERROR] cannot hash $dir/$f" >&2; exit 1; }
             printf '%s  %s\n' "$inner" "$f"
         done
-    ) | { if command -v sha256sum > /dev/null 2>&1; then sha256sum; else shasum -a 256; fi; } | cut -d' ' -f1
+    ) || return 1
+    # An empty tree hashes as EMPTY INPUT (python's tree_digest of no files);
+    # printf '%s\n' of an empty capture would inject one newline.
+    { [ -n "$listing" ] && printf '%s\n' "$listing" || printf ''; } | { if command -v sha256sum > /dev/null 2>&1; then sha256sum; else shasum -a 256; fi; } | awk '{ sub(/^\\/, "", $1); print $1; exit }'
 }
 
 # Refuse to build an unverified plugin tree.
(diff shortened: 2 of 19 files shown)


webserver/debug_websocket.py
```diff
@@ -7,16 +7,64 @@
...
+
+    The connect handler authenticates once. Without this, an open socket keeps
+    answering commands after its token expires (15 minutes by default -- the
+    config sets no JWT_ACCESS_TOKEN_EXPIRES) and after /logout revokes it (the
+    blacklist is only consulted by verify_jwt_in_request). That matters here
+    because the license FCs read the hardware anchor and write the license blob:
+    this channel is a trust boundary, so "authenticated once, ever" is not
+    enough. Re-verifying per command is cheap -- an HMAC and a set lookup.
+    """
+    token = _session_tokens.get(request.sid)
```
thiagoralves, line 39: **Re-verifying the connect-time token per command kills every long-lived debug session after ~15 minutes, because nothing ever hands this socket a fresher token.**

`handle_connect` stores the connect-time token (line 136) and this function replays *that* token on every subsequent command. Nothing in the runtime sets `JWT_ACCESS_TOKEN_EXPIRES` (grep across `webserver/` finds no occurrence), so flask-jwt-extended's 15-minute default applies — and there is no `refresh_token` event on this namespace for a client to push a new one.

The editor cannot work around it either: `openRestSession` in openplc-editor builds the WebSocket candidate as `create: () => new WebSocketDebugTransport({ host, port: 8443, token, … })`, capturing `config.connectionParams.jwtToken` in a closure at login time, and the transport connects with `reconnection: false`. So the token this session presents is frozen for the life of the session.

Concrete scenario: the user logs in, starts the debugger, and watches variables. At T+15min `verify_jwt_in_request()` raises `ExpiredSignatureError`; from then on **every** `debug_command` on the still-open socket answers `{"success": false, "error": "Unauthorized"}`. The editor's poll sees a generic failure it cannot distinguish from a dropped link, and the licence FCs this PR adds fail the same way — the badge reports `check-failed` on a healthy, reachable device (and openplc-editor#1023's `quietCheckFailed: true` means the user is not even told why).

The interval also can't be papered over by re-opening the channel: a newly created WebSocket reuses the same frozen token, so its `connect` is refused too.

Suggestion: keep the per-command re-verification (it is the right call for a trust boundary) but pair it with a way to renew — e.g. accept a `reauth` event carrying a fresh access token that replaces `_session_tokens[request.sid]`, and emit a distinguishable `token_expired` error so the client knows to renew rather than treating it as a transport failure. That needs a matching change on the editor side.

webserver/vpp_license_debug.py
```diff
@@ -0,0 +1,425 @@
...
+    if not data or data[0] not in _LICENSE_FCS:
+        return None
+    fc = data[0]
+
+    if fc == FC_GET_BOARD_ID:
+        anchor = _read_anchor()
+        if not anchor:
+            # Match the Arduino firmware (D57): no id -> SUCCESS with id_len=0, so
+            # the editor sees a clean empty id (outcome no-id), not an error byte.
+            return _hex_from_bytes(bytes([fc, ST_SUCCESS, 0]))
```
thiagoralves: **Answering an unreadable anchor with `SUCCESS, id_len = 0` hands the editor a licensing identity that belongs to no hardware.**

The Arduino convention this mirrors is correct for the readiness-probe role of 0x48: a core built without `ArduinoUniqueID` really is "OpenPLC firmware, no unique id". But on this medium 0x48 is *only* the licensing anchor — the module docstring says so, and openplc-editor#1023's `websocket-debug-transport.ts` calls it "the anchor read, not a readiness probe". For that role, zero bytes is not a valid answer.

What the pair actually does with it: `parseGetBoardIdResponse` returns `{ success: true, boardId: <0 bytes>, boardIdHex: '' }` for `id_len = 0`; the editor's `readLicenseAnchor` then does `board.boardId ?? new Uint8Array(0)` and proceeds, so `deriveDeviceId` hashes an empty pre-image. Every device that answers this way derives the **same** deviceId — and the closed core, which reads `/proc/device-tree/serial-number` itself, never reproduces it.

Concrete scenario: a runtime-v4 install on hardware without a device-tree serial (x86 industrial PC, a container, or a Pi whose `/proc/device-tree` is not mounted). `_read_anchor()` returns `b""`, 0x48 answers SUCCESS/0, the editor shows a device id, the user completes a purchase bound to that id, 0x49 writes the blob — and the plugin refuses it with DEVICE_MISMATCH and runs the 2-hour demo forever. That is precisely the failure the anchor-parity work exists to prevent, and the same code path also refuses `> 64` bytes with an error byte rather than returning something unusable.

Answer an error status here instead (`ST_LIC_UNSUPPORTED` reads naturally: this device has no anchor to license against), so the editor reports `check-failed` rather than inventing an identity. openplc-editor#1023 should refuse a zero-length anchor on its side too — there is an open bot comment on `readLicenseAnchor` for that half.

scripts/compile.sh
```diff
@@ -144,11 +234,32 @@ if [ -d "$VPP_PLUGIN_DIR" ] && [ -f "$VPP_PLUGIN_DIR/Makefile" ]; then
...
         echo "[INFO] VPP plugin compiled successfully"
     fi
+
+    # Record the sha256 of every .so this verified build produced, so the
+    # plugin loader can refuse an object swapped in AFTER the compile
+    # (core/src/drivers/vpp_plugin_seal.c, checked immediately before dlopen).
+    # Written on the cache-hit path too: the cached .so belongs to this same
+    # verified tree, and a runtime upgraded onto an existing build/vpp/ would
+    # otherwise have no seal at all and refuse to load a legitimate plugin.
+    : > "$VPP_OBJECT_SEAL"
```
thiagoralves, line 287: **Re-sealing on the cache-hit path voids the guarantee the seal exists for.**

`vpp_plugin_seal.h` states the purpose plainly: "an object swapped into `build/vpp/` AFTER the compile would load unchecked". But this block runs unconditionally — it is outside the `if [ "$NEEDS_COMPILE" -eq 1 ]` guard — so on a cache hit no `make` runs and the loop simply hashes and blesses whatever `lib*_plugin.so` happens to be sitting in `$VPP_OUTPUT_DIR` at that moment.

Concrete scenario: a swapped `.so` in `build/vpp/` is correctly refused by `native_plugin_get_symbols` on the next start (hash mismatch). The operator then re-uploads the *same* program from the editor. `check_vpp_verification_seal` passes (the `vpp_plugin/` tree is unchanged and still signed), the checksum cache hits, `NEEDS_COMPILE=0`, and this loop writes the swapped object's own hash into `vpp_plugin.seal`. It now loads. One ordinary re-upload converts a detected tamper into a blessed one, permanently.

The comment above justifies the unconditional write as covering "a runtime upgraded onto an existing `build/vpp/`", which is a real case — but it can be served without re-blessing unknown bytes, e.g. only write the seal when `NEEDS_COMPILE=1`, and on the cache-hit path only when `$VPP_OBJECT_SEAL` is absent *and* the cached checksum matched (so the objects provably belong to the tree just verified).

scripts/compile.sh
```diff
@@ -109,21 +109,111 @@ make -j"$JOBS" -f scripts/Makefile.strucpp
...
+# for every regular file, sorted by path. Must stay byte-identical to
+# webserver/vpp_package_signature.py's tree_digest(), which is what produced
+# the value in the seal -- hence LC_ALL=C for byte-order sort, POSIX relative
+# paths, and exactly two spaces.
+vpp_tree_digest() {
+    local dir="$1" f inner
+    (
+        cd "$dir" || exit 1
+        find . -type f -print | sed 's|^\./||' | LC_ALL=C sort | while IFS= read -r f; do
+            inner=$(sha256_hex "$f") || exit 1
```
thiagoralves: **`vpp_tree_digest` cannot report a hashing failure, so a bad digest is misdiagnosed as tampering.**

Two problems in this pipeline:

1. `|| exit 1` exits the `while` subshell (it is a pipeline stage), not `vpp_tree_digest`. The function's exit status is that of the last stage — `cut` — which is 0. So `if ! actual=$(vpp_tree_digest "$VPP_PLUGIN_DIR") || [ -z "$actual" ]` can never fire: on failure it produces a non-empty sha256 of a *partial* file listing.
2. `sha256_hex` takes the first space-delimited field of `sha256sum` output, but GNU `sha256sum` prefixes the whole line with a literal `\` and escapes the name when the filename contains a backslash or newline. `cut -d' ' -f1` then yields `\<hash>` rather than `<hash>`.

Either way the build still fails closed, which is the important part — but it fails with `"[ERROR] The VPP plugin tree changed after it was verified"` and a digest diff, telling the operator their signed package was tampered with when the real cause is a missing coreutils or an unusual filename. That message will send someone re-downloading a package that was fine.

Fix: capture the loop's failure explicitly (e.g. write to a temp file and check the writer's status, or `set -o pipefail` around the pipeline) and read the hash with `awk '{print $1}'` on a line whose leading `\` has been stripped. `webserver/vpp_package_signature.py:tree_digest` — the side that produced the sealed value — has neither problem, so any divergence here is a false accusation.

core/src/drivers/plugin_config.c
```diff
@@ -21,7 +21,65 @@ static void remove_newline(char *str)
...
+            {
+                return 0;
+            }
+        }
+        cursor++;
+    }
+
+    if (require_contained)
+    {
+        if (path[0] == '/' || path[0] == '\\')
```
thiagoralves, line 67: **The C and Python containment rules disagree on absolute paths, so one class of conf is accepted by the webserver and then silently dropped by the loader.**

`validate_vpp_plugins_conf` in `webserver/plcapp_management.py` deliberately tolerates an absolute `path` — `against_root()` branches on `os.path.isabs(candidate)` — and only requires that it resolve inside `build/vpp/`. This function rejects **every** absolute path when `require_contained` is 1, which is how `plugin_driver_append_config` now parses `vpp_plugins.conf`.

Concrete scenario: a conf carrying `/opt/OpenPLC_v4/build/vpp/libfoo_plugin.so` (same file the relative form names). The upload is accepted, `vpp_plugins.conf` is installed, `apply_vpp_plugin_conf` copies the config JSON and any `.license` beside it, `compile.sh` builds the `.so` and seals it — and then the entry is dropped at parse time with a `[PLUGIN] rejected plugin …` line in the runtime log while the editor reports a completely successful upload. The VPP simply never loads, and nothing in the upload response says why.

Today's editor emits `./build/vpp/lib<name>_plugin.so`, so this is latent rather than live — but the two guards are documented as mirroring each other ("Mirrors the Python-side containment in webserver/plcapp_management.py") and they do not. Either tighten the Python side to reject absolute paths too, or have this side accept an absolute path that resolves inside the runtime tree; whichever, the two should agree so a conf is never half-accepted.

thiagoralves (commented):
Correctness review of the runtime half, read together with openplc-editor#1023 and openplc-web#681. Five findings below, ordered roughly by impact: the per-command JWT re-verification has no renewal path (breaks debugging and licensing 15 minutes after login), the empty-anchor answer feeds the editor an identity no device reproduces, and the cache-hit re-seal undoes the loader seal's stated purpose. The signature-gate design itself (Ed25519 over the canonical payload, both directions of the present/missing file comparison, the atomic license write, and the `is_inside_root` unification) reads correct to me.

marconetsf:
All five findings addressed in `c96b374` (commit message maps each): R1 gains the `reauth` event + distinguishable `token_expired` (editor half in openplc-editor#1023 `349919137`: candidate reads the token manager at create(), held channel gets renewals pushed); R2 answers LIC_UNSUPPORTED for an unreadable anchor; R3 seals only what this run compiled and forces a recompile when the seal cannot vouch for cached objects (your suggested variant still blessed unknown bytes in the no-seal upgrade case — rebuilding from the just-verified tree closes it fully); R4 propagates hashing failures and strips sha256sum's escape marker (empty tree = empty input, matching python); R5 rejects absolute paths webserver-side, mirroring the C for real. Full pytest 161 green.

thiagoralves (approved):

PR #170, opened 2026-08-24
refactor(upload): drop the VPP package-signature verification on the upload path

marconetsf:
## Why

The runtime-side VPP package-signature check was half of a contract whose editor half was never built: it demanded a `vpp_signature.json` sidecar that **no released editor forwards**, so every licensed runtime-v4 upload died at *"vpp_signature.json is missing or unreadable"* the first time one was tried on real hardware. It never functioned end to end in the field. Rather than implement the missing editor half, the incomplete scaffolding is removed from the runtime — provenance of an installed VPP is already verified editor-side (the editor checks the package's Ed25519 signature before it builds the upload bundle).

This did not sit on a privilege boundary. An authenticated uploader already obtains root code execution through the ordinary program path (`scripts/Makefile.strucpp` compiles `core/generated/*.cpp`, including `c_blocks_code.cpp`, into `new_libplc.so`, which the root runtime dlopens). Removing this check opens no door that was otherwise shut.

## What

**Removed**
- `webserver/vpp_package_signature.py` (Ed25519 verify, sidecar parse, tree digest, verification seal) and its wiring in `webserver/app.py`;
- the `vpp_plugin.verified` interlock in `scripts/compile.sh` (`VPP_VERIFIED_SEAL`, `vpp_tree_digest`, `check_vpp_verification_seal`);
- tests: `test_vpp_plugin_signature.py`, `test_vpp_tree_digest_cross_language.py`;
- `cryptography` from `requirements.txt` — a leftover of the deleted module (see below).

**Kept, behaviour unchanged**
- the dlopen **object seal** (`build/vpp/vpp_plugin.seal`, `vpp_plugin_seal.c/h`, the compile.sh write/cache logic): it binds what the loader executes to what THIS runtime's compile step built — a distinct concern (post-compile object swap), unaffected by this change;
- the whole licence delivery surface (FCs 0x48–0x4B, `vpp_license_debug`).

The second commit erases the comment/dependency traces the first left behind (three stale comments and the `cryptography` dep).

## On `cryptography`

Removed, not kept. History: it was dropped from `requirements.txt` when certificate generation moved to the OpenSSL CLI (`980fbb2`) and re-added **only** by the signature feature (`2da6c26`). The sole `from cryptography` in the runtime-venv scope was the deleted module; the OPC UA security profiles that also use it run in their own isolated `venvs/opcua`. Proven: `pip uninstall cryptography` in the runtime venv, suite still green.

## Proof

- `bash -n scripts/compile.sh` clean; `py_compile webserver/app.py` clean; `git grep` for every removed symbol + `signature-verified` / `no provenance` returns zero hits.
- pytest (runtime venv, `cryptography` uninstalled): **122 passed, 16 skipped** (skips are the anchor cross-language tests that need the sibling openplc-packages checkout — environmental).

**Recovery**: the full prior implementation is preserved on branch [`archive/vpp-plugin-signature-gate`](https://github.com/Autonomy-Logic/openplc-runtime/tree/archive/vpp-plugin-signature-gate).

https://claude.ai/code/session_015uUH3ZL5ehreMUf2dtanWD


Gustavohsdp:
## Code review — what needs to change before merge

**Verdict: request changes.** Nothing conceptual — the product decision is sound and well recorded. What's missing is completeness of the removal. I estimate 30–60 minutes of work.

I'm not relitigating the decision itself: it's dated (2026-08-26), justified by a bench test, the trade-off is written down honestly, and the recovery branch `archive/vpp-plugin-signature-gate` **actually exists on origin** (I checked). That's above average for a security-control removal.

**What I verified myself:** `bash -n scripts/compile.sh` and `python3 -m py_compile webserver/app.py` are clean; the branch contains current `development` (`f7dce88`); and `git grep` across the whole tree for `vpp_package_signature`, `verify_uploaded_vpp_plugin`, `write_verification_seal`, `VPP_VERIFIED_SEAL`, `check_vpp_verification_seal`, `vpp_tree_digest`, `signature_required`, `vpp_signature.json` and `treeDigest` returns **zero** hits. The code decoupling is genuinely clean.

**What I could not verify:** the `122 passed, 16 skipped` claim. My local Python is 3.9.6 and `pyproject.toml` requires `>=3.10`. Worth noting that CI can't verify it either — the only workflows are `docker.yml` and `windows-installer.yml`, neither runs pytest.

---

## Blockers

### 1. Three orphan comment references, against the PR's own "zero orphan references in the tree"

The **Proof** section claims zero orphan references. That's true for *code* — I confirmed it. It isn't true for the comment narrative, and this PR deliberately rewrote that narrative in two files, so the standard is the one the PR set for itself.

**1a. `core/src/drivers/plugin_driver.c:1348-1353`** — the important one. It describes the removed gate as if it were still there, at the exact call site of the seal:

```c
    /* Last metre before execution: a VPP plugin .so must match the hash
     * scripts/compile.sh sealed after building it from a signature-verified
     * upload. The upload gate proved the plugin's INPUTS came from a signed
     * package; it cannot speak for the .so, which is linked here afterwards.
     * Without this check an object dropped into build/vpp/ after the compile
     * would be dlopen'ed with no provenance at all.
```

*Why it matters:* this is the sibling file of `vpp_plugin_seal.h`, whose header you carefully rewrote here (lines 5-13). Whoever opens `plugin_driver.c` first walks away believing an upload signature gate exists. And the closing line ("would be dlopen'ed with **no provenance at all**") only parsed when there *was* provenance — today there is none in either branch of that sentence.

Suggested, matching the wording already approved in `vpp_plugin_seal.h`:

```c
    /* Last metre before execution: a VPP plugin .so must match the hash
     * scripts/compile.sh sealed when it built that .so on this device.
     * The check binds what the loader executes to what THIS runtime's
     * compile step produced, so an object dropped into build/vpp/ after
     * the compile is refused. It says nothing about where the uploaded
     * sources came from -- package provenance is the editor's concern
     * (see vpp_plugin_seal.h, PR #170).
```

**1b. `scripts/compile.sh:173`** — same hunk you already touched. Line 178 was fixed (`"recompiling from the verified tree"` → `"from the uploaded tree"`), line 173 was not:

```bash
   172	                # upgraded runtime with no seal must REBUILD from the
   173	                # just-verified tree, never bless unknown bytes.
```

→ `just-extracted tree`. One word.

**1c. `tests/pytest/plugins/test_vpp_license_delivery.py:83`** — points at a file this PR deletes:

```python
    # .so to resolve inside build/vpp/ (see test_vpp_plugin_signature.py). A fake
```

Redirect to wherever `validate_vpp_plugins_conf` is actually covered, or drop the parenthetical.

*Why this blocks rather than being cosmetic:* the PR makes a verifiable claim in **Proof** and it doesn't hold. For a PR whose entire value is "I removed a security control cleanly and completely", the auditability of that claim **is** the deliverable. Either fix the three comments, or narrow the Proof line to "zero orphan code references".

---

### 2. Zero test coverage for the behaviour this PR exists to fix

Test balance: **−701 lines, +0**.

The stated reason for the PR is that *every licensed runtime-v4 upload died at "vpp_signature.json is missing or unreadable"*. There is now **no test that fails if that bug comes back**.

This is a direct miss against `.claude/review-guidelines.md`:

> *"New behavior comes with pytest coverage where the Python side is touched."*

`webserver/app.py` was touched. The new behaviour is one assertion: **an upload carrying `vpp_plugin/` with no signature sidecar now compiles**. That's the only guard against someone reintroducing a gate — or a stray import — and silently breaking the licensing path in the field again.

Suggested (`tests/pytest/test_vpp_upload_no_signature.py`):

```python
def test_should_accept_vpp_upload_without_any_signature_sidecar(monkeypatch, client):
    """A vpp_plugin/ upload with no vpp_signature.json must reach COMPILING.

    Regression guard for PR #170: the runtime-side package-signature gate was
    removed because no released editor forwards the sidecar. If a gate is ever
    reintroduced without the editor half, this test fails instead of the field.
    """
    # Arrange
    zip_bytes = make_upload_zip({
        "generated.hpp": b"...",
        "configuration.cpp": b"...",
        "vpp_plugin/Makefile": b"all:\n\t@true\n",
        "vpp_plugin/checksum.sha256": b"deadbeef  Makefile\n",
    })  # deliberately NO vpp_signature.json
    monkeypatch.setattr(plcapp_management, "run_compile", lambda *a, **k: None)

    # Act
    resp = client.post("/api/upload-file",
                       data={"file": (io.BytesIO(zip_bytes), "p.zip")},
                       headers=auth_header())

    # Assert
    assert resp.status_code == 200
    assert resp.get_json()["UploadFileFail"] == ""
    assert resp.get_json()["CompilationStatus"] == BuildStatus.COMPILING.name
```

One test. It's the only artefact in this PR that outlives the memory of the person who wrote it.

---

### 3. The security trade-off lives only in the PR description; `docs/SECURITY.md` is untouched

You removed a security control and documented the accepted risk clearly — in a place that **disappears from view the moment this merges**. Nobody audits PR descriptions six months later.

`docs/SECURITY.md` has a **"File Upload Security"** section (lines 110-180) documenting exactly this surface: path traversal, size limits, zip-bomb detection, `DISALLOWED_EXT`, safe extraction. It's the canonical home of this threat model, and it is now **incomplete by omission** — it describes the upload path's defences without mentioning that VPP subtree provenance verification was deliberately dropped.

Repo guideline: *"Docs updated when documented behavior changes (README, CLAUDE.md, docs/)."*

Suggested addition at the end of "File Upload Security":

```markdown
### VPP Plugin Package Provenance

**Status: not verified runtime-side (decision 2026-08-26, PR #170).**

An upload carrying a `vpp_plugin/` subtree is compiled by `scripts/compile.sh`,
which runs the uploaded `Makefile` with the runtime's privileges (`User=root`,
see `install.sh`). The runtime does **not** check who produced that subtree.

Provenance is checked by the OpenPLC Editor, which verifies the installed VPP's
Ed25519 signature before it builds the upload bundle. A hand-crafted upload sent
straight to `POST /api/{command}` bypasses that check entirely.

**Residual controls on this path**
- `@jwt_required()` on the endpoint — an authenticated session is required.
- `analyze_zip` / `safe_extract` — path traversal, size, and zip-bomb checks.
  Note `DISALLOWED_EXT` is a **denylist** and does not cover extensionless
  files, so `vpp_plugin/Makefile` passes by design.

**What this is NOT mitigated by:** the dlopen object seal
(`build/vpp/vpp_plugin.seal`) does not apply here — see note below.

**Accepted because:** an authenticated uploader can already obtain code
execution as root through the ordinary program path — `scripts/Makefile.strucpp`
compiles `$(wildcard core/generated/*.cpp)`, including `c_blocks_code.cpp`, into
`new_libplc.so`, which the root runtime dlopens. The VPP gate did not sit on a
privilege boundary that was otherwise closed; it enforced licence integrity.

**Prior implementation preserved on:** `archive/vpp-plugin-signature-gate`.
```

> On that last paragraph — I chased the worst case (`vpp_plugin/Makefile` executed as root via `make -C`; `Makefile` slips past `DISALLOWED_EXT` because it has no extension; `install.sh:117` confirms `User=root`). **It is not a new escalation.** An authenticated uploader already gets root code execution through the normal program path. Removing the gate does not open a door that was otherwise shut — it removes licence enforcement. **Your framing in the PR body is correct**, and it's worth writing that reasoning into `SECURITY.md` so the next person doesn't have to redo the investigation.

---

## Non-blocking, but worth fixing here

**4. `requirements.txt:8-10` — the new justification names the wrong environment.**

```
# Certificate/key handling for the OPC UA server security profiles
# (core/src/drivers/plugins/python/opcua/opcua_security.py).
cryptography
```

Keeping the dep instead of dropping it blindly was the right instinct, but the reason doesn't survive the venv topology:

- Root `requirements.txt` installs into **`venvs/runtime`** (`install.sh:575-577`).
- `opcua_security.py` runs in **`venvs/opcua`**, built by `setup_plugin_venvs` (`install.sh:367-414` → `manage_plugin_venvs.sh`) from the plugin's own `requirements.txt` — where `cryptography` is **commented out** and arrives transitively via `asyncua==1.1.8`.
- On Linux that venv is plain `python3 -m venv`, no `--system-site-packages` (`manage_plugin_venvs.sh:129-137` — the flag is MSYS2-only). The two environments are isolated.
- `pyproject.toml` declares no `dependencies`.

So `opcua_security.py` never imports the `cryptography` this line installs. And since it's the only `from cryptography` in the tree, the dep now looks unused in `venvs/runtime` (Flask-JWT-Extended / PyJWT default to HS256, which doesn't need it).

Suggest: `pip uninstall cryptography` in `venvs/runtime` + run the suite. Green → drop the three lines. Something breaks → name the real consumer. A dependency kept for a provably false reason is worse than one with no comment: it invites the next person to trust it.

**5. "The remaining mitigation is the dlopen object seal" — category error.**

The seal doesn't cover the vector this PR opens. Walking `scripts/compile.sh:161-215` for a hand-crafted hostile upload:

1. `vpp_plugin/` with a `Makefile` → enters the block (line 161)
2. no cache hit → `NEEDS_COMPILE=1` → `make -C "$VPP_PLUGIN_DIR"` runs the hostile Makefile
3. **`: > "$VPP_OBJECT_SEAL"`** — the seal is rewritten from scratch over exactly what was just produced
4. the loader checks it and accepts

The seal seals the attack's output. It defends against **post-compile object swap** — a different threat, one that presupposes filesystem write. And even there its value is bounded: `build/vpp/vpp_plugin.seal` is unkeyed plaintext next to the `.so`, with no owner/permission check in `vpp_plugin_seal.c` (`#define VPP_SEAL_FILE VPP_BUILD_SUBDIR "/vpp_plugin.seal"`, line 15). Anyone who can swap the `.so` rewrites the seal.

To be fair: `vpp_plugin_seal.h:19-22` is already honest about this (*"Not a defence against someone with root and a text editor -- the seal is unkeyed"*). The issue is only the PR body promoting it to a mitigation of something it doesn't mitigate — and that sentence is what becomes the institutional memory of this decision.

Suggest: *"No runtime-side mitigation remains on this path; the dlopen object seal continues to cover a different threat (post-compile object swap) and is unaffected."*

**6. `scripts/compile.sh:104-118` — the header now warns what *isn't* a guarantee without saying what is.**

The paragraph you removed was the one that answered the question. What's left ends at line 116 with *"It proves nothing about where the plugin came from"* and then a bare `#`. The reader learns `checksum.sha256` proves nothing, and never finds out that today **nothing** does. In a script that runs an uploaded Makefile as root, that's the most important fact in the file. Replacing the dangling line 117:

```bash
# NOTHING in this script gates that provenance. The Makefile below comes from
# the upload and runs with this process's privileges (root, per install.sh).
# The runtime-side signature gate was removed by decision 2026-08-26 (PR #170);
# provenance is verified editor-side, before the bundle is built.
# See docs/SECURITY.md, "VPP Plugin Package Provenance".
```

**7. `core/src/drivers/vpp_plugin_seal.h:13` — durable reference anchored to a branch name.**

The branch exists, I checked. Still, a branch is the weakest possible anchor in a source comment: deletable by anyone with push, vanishes in housekeeping, leaves no trace once pruned. A PR number is immutable and always resolves. Suggest `preserved on branch archive/vpp-plugin-signature-gate (PR #170)` — branch as convenience, PR as anchor. If the repo does ADRs, this is a textbook candidate.

**8. PR template not followed.** `.github/pull_request_template.md` asks for four sections and a checklist. Your Why / What / Trade-off / Proof structure is honestly *more* informative — but it drops the items the template exists to force: **Ticket (`RTOP-<n>`)** is missing, `pre-commit run` clean isn't stated, "Docs updated if behavior changed" isn't done (see #3). Suggest keeping the narrative and appending the checklist under it.

---

## Nits

- `scripts/compile.sh:144-146` — three consecutive blank lines where `vpp_tree_digest()` and `check_vpp_verification_seal()` were removed. Collapse to one.
- `scripts/compile.sh:117` — the stray `#`. Goes away if you apply #6.
- Legacy `core/generated/vpp_plugin.verified` on already-deployed devices: harmless (nothing reads it, and `handle_upload_file` does `shutil.rmtree(extract_dir)` on every upload, so the first post-upgrade upload clears it). Flagging only so nobody later mistakes such a file for evidence the gate still runs.
- Process, not this PR: `docs/pr-reviews/PR_REVIEW_CHECKLIST.md` could use an explicit *"security control removed → record the accepted risk in `docs/SECURITY.md`, not just in the PR description"*. #3 happened because nothing in the process required it.
- Process, not this PR: no workflow runs `scripts/run-pytest.sh`. A pytest job would turn "122 passed" from a claim into a fact, and would give the test in #2 teeth. That's the root cause of #2 slipping through.

---

## Prior-review markers — all checked

There are no earlier reviews on this PR, so nothing to verify there. But there *are* prior-review markers recorded in the code this PR touches. I traced each one:

| Marker | Original fix | Post-PR | |
|---|---|---|---|
| **2026-08-20, R3** (a) | Cache hit only stands when the seal vouches for the `.so` on disk | `vpp_object_seal_matches()` intact (`compile.sh:147-159`), cache-hit guard intact (`compile.sh:174`) | ✅ preserved |
| **2026-08-20, R3** (b) | Write the seal only when `NEEDS_COMPILE=1`, never on cache hit | `if [ "$NEEDS_COMPILE" -eq 1 ]` guard intact before `: > "$VPP_OBJECT_SEAL"` | ✅ preserved |
| **2026-08-20, R3** (c) | Upgraded runtime with no seal must REBUILD, never bless unknown bytes | Logic intact; only the comment at `compile.sh:173` went stale | ⚠️ see 1b |
| **2026-08-20, R4** | `pipefail` in `vpp_tree_digest`'s subshell, so a hashing failure couldn't emit a partial-listing digest and accuse an intact package of tampering | Removed together with the function | ✅ correct — the fix was scoped to that function |

No prior remediation was lost by accident. R4 went away as a legitimate consequence of deleting the function it fixed; R3 survives intact in the logic. The only damage is to R3's comment at `compile.sh:173`.

---

## What's good here — worth naming

1. **The code decoupling is genuinely clean.** Zero orphan imports, calls, file reads, or references in Dockerfile / install.sh / pre-commit. Only the comments in #1 slipped.
2. **You wrote down the security trade-off before anyone asked.** Cost quantified ("minutes, low skill" → "fork and reinstall the runtime"), the new holder of the guarantee named, the residue stated. That's rare and it's exactly right when removing a control.
3. **The recovery plan is real, not rhetorical** — I verified the archive branch on origin.
4. **Surgical scope.** One commit, 8 files, and the explicit decision *not* to touch the object seal, the licence delivery surface (FCs `0x48-0x4B`), or `vpp_license_debug`. No drive-by refactors.
5. **Operator-facing strings updated.** The three `log_error` calls at `vpp_plugin_seal.c:262/301/309` no longer name a gate that doesn't exist. That's the detail that almost always gets left behind — and it's what a field technician reads at 2am.
6. **`cryptography` kept rather than reflex-deleted.** Right call; the justification just needs fixing (#4).
7. **`vpp_plugin_seal.h` rewritten honestly.** The new header narrows the claimed guarantee to exactly what survives, without inflating it. It's the behaviour #1a and #5 only need to propagate to the other two places.

---

## Summary

| # | Action | File | Est. |
|---|---|---|---|
| 🔴 1a | Rewrite the removed-gate comment | `core/src/drivers/plugin_driver.c:1348-1353` | 5 min |
| 🔴 1b | `just-verified tree` → `just-extracted tree` | `scripts/compile.sh:173` | 1 min |
| 🔴 1c | Drop ref to deleted test file | `tests/pytest/plugins/test_vpp_license_delivery.py:83` | 1 min |
| 🔴 2 | Regression test: VPP upload without sidecar → `COMPILING` | new file in `tests/pytest/` | 20 min |
| 🔴 3 | Record the accepted risk under "File Upload Security" | `docs/SECURITY.md` | 15 min |
| 🟡 4 | Fix or drop the `cryptography` justification | `requirements.txt:8-10` | 5 min |
| 🟡 5 | Correct the object-seal sentence | PR body | 2 min |
| 🟡 6 | State what does *not* gate the block | `scripts/compile.sh:117` | 3 min |
| 🟡 7 | Anchor to PR #170, not just the branch | `core/src/drivers/vpp_plugin_seal.h:13` | 1 min |
| 🟡 8 | RTOP ticket + template checklist | PR body | 3 min |
| 🟢 | Blank lines and the stray `#` | `scripts/compile.sh:117,144-146` | 1 min |

With **1–3** resolved I'm happy to approve. **4–8** I'd take as non-blocking if the team wants to unblock the merge — except **4**, which I'd rather see fixed now, because a dependency justified by a false reason tends to survive for years.

One last note, about process rather than code: you put real effort into honestly documenting a security trade-off, and that effort went into the most volatile field on GitHub. Item #3 isn't a complaint about the PR — it's about the repo not having a place that forces such a record to outlive the merge. The `PR_REVIEW_CHECKLIST.md` nit above addresses the cause. Worth doing.

marconetsf:
@Gustavohsdp thank you — the trace-hunting was exactly right, and your root-exec investigation reframed the whole thing. Follow-up in `6134327`.

A reframing first, because it decides several items: **treat this as removing incomplete scaffolding that never functioned, not as removing a production security control.** Your own analysis is what licenses that — you proved the check never sat on a privilege boundary (an authenticated uploader already gets root via `Makefile.strucpp` → `new_libplc.so`), and the editor half of the sidecar contract was never built, so the gate never worked end to end in the field. Under that lens the goal is to erase the traces, not to document an accepted risk. So the completeness items are in; the "record the residual risk" items are declined, with reasons below.

**Done (the genuine leftovers you caught):**
- **1a** — `plugin_driver.c` rewritten to describe only what the object seal does; dropped "signature-verified upload" and "no provenance at all". I diverged from your suggested text in one way: no `PR #170` anchor in the comment (see 7).
- **1b** — `just-verified` → `just-extracted`.
- **1c** — dropped the pointer to the deleted `test_vpp_plugin_signature.py`.
- **4** — you were right, and further than "fix the comment": the dep is a leftover. It was dropped when cert-gen moved to the OpenSSL CLI (`980fbb2`) and re-added only by the signature feature (`2da6c26`); the OPC UA venv is isolated. **Removed the dependency**; `pip uninstall cryptography` in the runtime venv, suite still 122/16.
- **6 (cleanup half)** — the dangling header and stray `#` are gone; the header now states what `checksum.sha256` IS (a cache key). Blank block collapsed.
- **7 (inverted)** — rather than anchor the header to `PR #170`, I **removed** the archaeology sentence entirely. A source comment narrating a removed feature is itself a trace; git history and this PR are where that belongs.

**Declined, with reasoning:**
- **2 (regression test)** — a test asserting "upload without a sidecar compiles" reintroduces the `sidecar` vocabulary as a monument to a concept we're erasing, and the vpp_plugin upload happy path is already exercised by `test_vpp_license_delivery.py`. There's no gate left whose reintroduction this would guard against without also reintroducing the editor half.
- **3 (`SECURITY.md` accepted-risk section)** — this is the one I'd push back on hardest, and it rests on your own finding: there is **no risk delta** to record. The check never gated a privilege boundary. A "VPP Plugin Package Provenance: not verified" section would imply a real control was dropped and send the next reader hunting for a boundary that never existed. Documenting a non-event is worse than silence. (The pre-existing "upload runs a Makefile as root" fact is true independent of this PR and could be its own SECURITY.md note if you want one — but that's not this change.)
- **5 (object-seal sentence)** — agreed the old sentence was a category error; it's gone with the body rewrite rather than corrected in place. The body no longer claims any residual mitigation on this path.
- **6 (the "NOTHING gates provenance / root / SECURITY.md" narrative)** — declined for the same reason as 3: it documents a non-control at the most visible spot in the file.
- **8 / checklist / PR_REVIEW_CHECKLIST line** — the RTOP card and the "security control removed → SECURITY.md" process rule both presume the security-control framing we're setting aside, so no card. The CI-runs-pytest gap you flagged is real but pre-existing and out of scope here.

On process: `pre-commit` I could only verify piecemeal (it's not installed in my shell, and per your own note no workflow runs it) — trailing-whitespace, final-newline and the Python/shell edits are clean by manual check; the C edits are comment-only and preserve the existing formatting. Body rewritten to match the reframing.

Gustavohsdp (approved):
Approving. The completeness items are done, and I verified your reframing independently rather than taking it on argument — it holds, and with stronger evidence than the PR offers.

**Trace sweep is clean.** `git grep` for every removed symbol plus `signature-verified` and `no provenance` returns zero. The only surviving mention is `vpp_plugin_seal.h:11`, describing what the *editor* does — still true, correct to keep. `bash -n` and `py_compile` clean; no blank-line runs left in `compile.sh`.

**The `cryptography` removal checks out commit by commit.** `980fbb2` ("Replace Python Cryptography with OpenSSL CLI for certificate generation") dropped it; `2da6c26` (the signature feature) re-added it. Removing it now restores the pre-feature state. Nothing else in the runtime venv needs it: TLS uses stdlib `ssl` with a cert file (`app.py:415-416`), cert generation shells out to the OpenSSL CLI (`credentials.py:118,152,211`), the only remaining `from cryptography` is `opcua_security.py` in its own venv, and MSYS2 is unaffected — `install.sh:315` still installs `python-cryptography` via pacman. Going further than "fix the comment" was the right call.

**On the reframing — I checked the editor repo, and you are right.** Three things settle it:

1. The editor's signature file is `signature.json` (`verify-package-signature.ts:24`). `vpp_signature.json`, the name the runtime demanded, appears **nowhere** in the editor.
2. The test that enumerates the upload bundle (`pipeline.test.ts:737-740`) lists `vpp_plugins.conf`, `vpp_plugin/Makefile`, `vpp_plugin/checksum.sha256` — no signature file of any kind.
3. Editor-side verification is real and sits on the compile path: `compiler-module.ts:2079,2601,3113` → `verifyBoardPackageIntegrity` → `verifyPackageSignature(packagePath, TRUSTED_PACKAGE_KEYS)` (`package-manager-module.ts:263`).

So the gate rejected 100% of `vpp_plugin/` uploads and provenance genuinely is checked editor-side. **I concede item 3**: there is no risk delta to record, and my "document the accepted risk" framing was wrong — it assumed a working control was being removed.

**One correction for the record**, on the reason given for declining item 2:

> the vpp_plugin upload happy path is already exercised by `test_vpp_license_delivery.py`

It isn't. That file has two tests — path derivation, and `apply_vpp_plugin_conf` in isolation against a hand-built fake. Neither touches `handle_upload_file`. I swept all 28 test files for `handle_upload_file|analyze_zip|safe_extract`; the only hit is `safe_extract_runtime_args_from_capsule` in the modbus test, unrelated. Nothing covers the upload path.

The decline itself is fine on scope — the gap is pre-existing and there is no gate left to guard — and the point about not reintroducing "sidecar" vocabulary has merit (though a test wouldn't need it; "an upload carrying `vpp_plugin/` reaches COMPILING" says it without the word). I'm only flagging that the coverage cited doesn't exist, so it doesn't stand as the record.

**Caveat on the proof:** `122 passed, 16 skipped` remains author-attested. I can't reproduce it — local Python is 3.9.6 against a `>=3.10` requirement — and this repo's CI doesn't run pytest (only `docker.yml` and `windows-installer.yml`). Everything else above I verified directly.

PR #183, opened 2026-09-01
RTOP-272: Retrieve Project from PLC (runtime)

thiagoralves:
Integration PR for the runtime's half of **Retrieve Project from PLC** (epic RTOP-272). Each child task was reviewed and merged into this branch individually; this is the branch going to `development`.

**This is the repo the other three depend on.** The editor and web clients are useless against a runtime without these endpoints, so this should land first.

A device now keeps a copy of the source project it is running, so the project can be opened again from a machine that does not have it — the case where an engineer left, a laptop died, or a panel has been running for two years and nobody knows which version is in it.

## Child PRs already merged here

| PR | Ticket | What |
| --- | --- | --- |
| #180 | RTOP-273 | Snapshot storage, the upload field, and the retrieval endpoints |
| #181 | RTOP-278 | Rescue a device whose accounts contain no administrator |
| #182 | RTOP-277 | Documentation, security first |

## What reviewers should look at

**The runtime never parses the archive.** It is an opaque blob; metadata arrives in separate form fields and is normalised into a record the device owns (unknown keys dropped, since everything there is echoed back to other clients). 100 MB cap.

**The lifecycle is the subtle part.** Three rules keep a device from ever advertising a project it is not running:

- The stored project is cleared at the upload's *point of no return*, not on arrival — so a rejected upload (bad zip, too large, runtime busy) leaves the previous program **and** its project untouched, which is the pair that is still true.
- The new project is staged only *after* `safe_extract` succeeds, and promoted or discarded in `run_compile`'s `finally`. Staging before the extract could strand a snapshot that the *next* successful build would then promote — belonging to an upload that never landed. That was a self-review catch and has a regression test.
- An upload carrying no project erases the stored one. Deliberate: `openplc-cli`, older editors and third-party clients upload without one, and a surviving project would describe a program that is gone.

**The discovery reply gained two fields**, `project_name` and `project_timestamp`, **absent rather than empty** when nothing is stored — which is what lets a client distinguish "nothing stored" from "stored but unnamed" without a separate flag. Note this reply is unauthenticated, so the project *name* is readable by anything on the network; the archive is not.

**Access control is one role check, and the docs say so plainly.** `GET /api/project-snapshot` is admin-only; `/info` is authenticated but not admin-gated, since it says nothing discovery does not already carry. The stored project is **not encrypted** and there is **no integrity guarantee** — `docs/RETRIEVE_PROJECT.md` and a new section in `docs/SECURITY.md` state this directly rather than softening it.

**The admin repair (#181)** exists because retrieval needs an administrator and nothing else did, so a device with no admin account was previously unnoticeable. A *sole* non-admin account is promoted at startup — it is already that device's whole administration. Several non-admin accounts are deliberately left alone: picking one would invent an authority the deployment never granted.

## Testing

65 tests across the snapshot, admin-repair, capabilities and discovery suites, all passing. Exercised end-to-end on real hardware via `openplc-cli`: upload with a project, retrieve it back, failed-compile discards, CLI upload clears.

**Pre-existing suite failures, not from this branch:** 14 failures (`test_vpp_anchor_cross_language.py`, `test_openplc_input_registers_datablock.py`) and 3 collection errors in the OPC-UA plugin tests (`pytest_asyncio` not installed in the test venv). All reproduce identically on `development` — verified by checking out `development` and running the same files.

## Cross-repo

- **openplc-editor** #1070 — the desktop client
- **openplc-web** #720 — the web client
- **orchestrator-agent** #125 (RTOP-281) — the discovery scan web needs for project names
- **RTOP-282** — the Edge relay endpoint, not started; until it ships the web picker lists devices without project names, behind a capability check

https://claude.ai/code/session_01Bpea15BYCgPSqpatpveRk1

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ The processes communicate via Unix domain sockets (`/run/runtime/plc_runtime.soc
 - [Editor Integration](docs/EDITOR_INTEGRATION.md) - How OpenPLC Editor communicates with the runtime
 - [Docker Deployment](docs/DOCKER.md) - Container usage and configuration
 - [Compilation Flow](docs/COMPILATION_FLOW.md) - How programs are built and loaded
+- [Retrieve Project from PLC](docs/RETRIEVE_PROJECT.md) - Storing the source project on a device and pulling it back
 - [Troubleshooting](docs/TROUBLESHOOTING.md) - Common issues and solutions
 
 ### Advanced Topics
--- a/docs/RETRIEVE_PROJECT.md
+++ b/docs/RETRIEVE_PROJECT.md
@@ -1,0 +1,223 @@
+# Retrieve Project from PLC
+
+A device keeps a copy of the source project it is running, so the project can be
+opened again from a machine that does not have it. Upload a program from the
+OpenPLC Editor or from openplc-web, and the source project travels with it;
+later, anyone with an administrator account on that device can pull it back.
+
+This answers the everyday case where the device is the only remaining record: an
+engineer left, a laptop died, a panel has been running for two years and nobody
+is sure which version of the project is in it.
+
+## Read this first: the stored project is not encrypted
+
+**The project is stored as a plain ZIP file on the device's filesystem. Anyone
+who can read that filesystem can read the project** -- over SSH, with physical
+access to the storage, from a backup image, or from a mounted volume in a
+container deployment.
+
+The access control on this feature is a single role check on the retrieval
+endpoint: the HTTP API will not hand the archive to a non-administrator. That is
+the whole of it. It protects the network path and nothing else.
+
+There is also **no integrity guarantee.** Nothing signs or checksums the stored
+project, so a project replaced on disk by someone with filesystem access is
+retrieved as if it were the original, and neither the device nor the client can
+tell. Do not describe this feature as verified, trusted, or tamper-proof,
+because it is none of those things.
+
+One more thing worth knowing: the **project name and its timestamp are
+advertised in the unauthenticated UDP discovery reply**, which is how a client
+can show what a device holds before anyone signs in. The name of a project is
+therefore readable by anything on the same network. The archive is not.
+
+If a deployment needs the project protected at rest, the answer is full-disk or
+filesystem-level encryption on the device. This feature does not provide it and
+is not a substitute for it.
+
+## What is stored
+
+The complete source project as the editor holds it -- POUs, data types, global
+variable lists, device and pin configuration, the project manifest -- plus a
+copy of every library the project references.
+
+The `build/` directory is not included. It is generated output, it is the
+largest part of a project, and it is reproduced by compiling.
+
+The archive is capped at 100 MB (`MAX_SNAPSHOT_BYTES` in
+`webserver/project_snapshot.py`). An upload carrying a larger one is accepted as
+a program upload, but the project is not stored, and the response says so.
+
+It lives beside the runtime's other persistent state, in `project_snapshot/`
+under the persistent data directory:
+
+| Deployment | Path |
+| --- | --- |
+| Native Linux | `/var/lib/openplc-runtime/project_snapshot/` |
+| Container | `/var/run/runtime/project_snapshot/` |
+| Overridden | `$OPENPLC_PERSISTENT_DATA_DIR/project_snapshot/` |
+
+## How it stays current
+
+The stored project always describes the program the device is actually running,
+which takes three rules working together.
+
+**Every upload replaces it.** When an upload reaches the point where the
+existing program is being overwritten, the stored project is erased. The new one
+is written only after the new program is safely in place.
+
+**An upload that carries no project erases the stored one.** This is deliberate,
+not an oversight. `openplc-cli`, older editors, and third-party clients upload
+programs without a source project; if the previous one survived, the device would
+advertise a project it is no longer running, which is worse than advertising
+nothing.
+
+**A failed compile leaves nothing behind.** The new project is held aside while
+the build runs and becomes the stored one only if the build succeeds. Any other
+outcome discards it. This matches what happens to the program itself: a failed
+build leaves the device with no program at all, so leaving a stored project
+would describe something that is not there.
+
+The upshot is that a device never advertises a project it is not running. It may
+advertise none -- that is the honest answer after a CLI upload or a failed build.
+
+## Retrieving
+
+The flow is the same on both clients: **File -> Retrieve Project from PLC**.
+
+1. **Pick a device.** The desktop editor scans the local network by UDP
+   broadcast; openplc-web asks each orchestrator's agent to run the same scan
+   against the runtimes it manages, so both read the same source of truth -- the
+   runtime's own advertisement. Each row shows the stored project's name and
+   timestamp. A device storing nothing is shown as such and cannot be chosen.
+
+2. **Give up the current connection, if it costs one.** The client holds a
+   session for one device at a time, so retrieving from a different device ends
+   the session you have. When that applies, you are told before it happens.
+   Picking the device you are already signed in to reuses that session and skips
+   straight to the retrieval.
+
+3. **Sign in as an administrator.** Any authenticated user can see *that* a
+   project is stored and what it is called; only an administrator can retrieve
+   the archive itself. A non-administrator who performed the upload still cannot
+   retrieve it.
+
+4. **The project opens.** It is unpacked into a scratch directory and opened as
+   an ordinary project.
+
+5. **Libraries are reconciled.** Each library the project carries is compared
+   against the ones on this machine by content hash, not by version number --
+   two different builds can carry the same version label. Libraries that are
+   missing, or present but different from what the project was built against,
+   are offered for installation. A library that is present and identical needs
+   nothing.
+
+### A retrieved project has no home yet
+
+It was unpacked into a scratch directory, not into a location anyone chose, so
+**Save is refused and points at Save As.** Saving in place would report success
+for work that is not where the user thinks it is, and on the desktop it would
+sit in a temporary directory. Use **File -> Save As** to give it a home; from
+then on it saves normally.
+
+Compiling a retrieved project works before saving it: the build's own internal
+flush is exempt from that refusal, because refusing it would not protect
+anything -- it would only stop the project from compiling.
+
+Save As is not yet implemented in openplc-web, where projects live in the user's
+account rather than on disk (tracked as RTOP-280). Until it is, retrieve on the
+web opens the project for inspection and compilation; keeping a copy means
+opening it in the desktop editor.
+
+## HTTP API
+
+Both endpoints require authentication. See [API.md](API.md) for the
+authentication flow.
+
+### `GET /api/project-snapshot/info`
+
+What is stored, without the archive. Authenticated, but not restricted to
+administrators: this is what a client needs to decide whether retrieving is
+worth offering, and it says nothing the discovery reply does not already carry.
+
+```json
+{
+  "present": true,
+  "projectName": "Traffic Light",
+  "editorVersion": "1.2.0",
+  "uploadedBy": "operator",
+  "timestamp": "2026-08-31T12:00:00Z",
+  "sizeBytes": 48213,
+  "formatVersion": 1,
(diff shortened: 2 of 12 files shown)


JulioSergioFS (changes requested):
# Review — openplc-runtime #183 · RTOP-272: Retrieve Project from PLC (runtime)

**Branch:** `RTOP-272-retrieve-project-from-plc` → `development` · 1312+/10− across 12 files
**Reviewed at:** integration branch head, against `origin/development`
**CI:** no checks configured on this branch — the whole verification burden is on the review and on the 65 tests described in the body.

**Verdict: request changes.** One memory-exhaustion path reachable by any authenticated non-admin, and one advertised capability flag that no client reads (which contradicts the description of two sibling PRs). Everything else is either a should-fix or a nit. The lifecycle design — clear at the point of no return, stage after extract, promote/discard in `run_compile`'s `finally` — is the strongest part of the epic and I could not break it on paper.

This is correctly identified as the repo that must land first.

---

## Blocking

### 1. The snapshot part is fully buffered before its size cap is checked

`webserver/app.py` — `stage_project_snapshot()`

```python
blob = snapshot_file.read()          # unbounded
...
project_snapshot.stage(blob, metadata)   # cap checked HERE, in project_snapshot.py
```

`MAX_SNAPSHOT_BYTES = 100 MB` is enforced inside `stage()`, i.e. **after** the whole part is already in memory. There is no `MAX_CONTENT_LENGTH` on the Flask app (I grepped `webserver/` — it is not set anywhere), so nothing bounds the part at the HTTP layer either.

The route is `POST /api/<command>` with `@jwt_required()` and **no admin gate**, so any account on the device — including a plain `user` — can post a multipart request with an arbitrarily large `snapshot` field. Werkzeug spools parts over ~500 KB to a temp file, so the sequence is: unbounded write to the device's disk, then `.read()` pulls all of it into RAM, then the cap rejects it. On the hardware this runtime targets, that is an unauthenticated-adjacent OOM/disk-fill.

The fix pattern already exists three lines away in the same function, applied to the program zip:

```python
if zip_file.content_length > MAX_FILE_SIZE:
```

Apply the same check to `snapshot_file` before `.read()`, and set `app.config['MAX_CONTENT_LENGTH']` as a backstop — `content_length` on a multipart part is not always populated by the client, so the check alone is not sufficient. A bounded chunked read (`read(n)` in a loop, abort past the cap) is the version that does not depend on a client-supplied header at all.

**Test gap that hid this:** `test_an_oversized_snapshot_is_refused` exercises `stage()` directly, so it proves the cap fires *after* the bytes exist. There is no test posting an oversized part through the upload endpoint. That is exactly the boundary the guard is supposed to defend.

### 2. `capabilities.projectSnapshot` is advertised but no client reads it

`webserver/restapi.py` — `restapi_capabilities()` gains `"projectSnapshot": True`.

I grepped both client repos on their PR branches:

```
grep -rn "projectSnapshot" openplc-editor/src openplc-web/src   # no hits outside tests
```

Neither #1070 nor #720 reads this flag. That matters because #1070's description says the editor *"degrades gracefully against a runtime without the feature (capability flag `projectSnapshot`)"* — it does degrade, but via the absence of `project_name` in the UDP advertisement, not via this flag. The flag as shipped is an unconsumed contract.

Either wire it in the clients (the honest place is the editor's retrieve modal, so a runtime without the feature says *"this device's runtime does not support retrieving projects"* rather than showing every row greyed out and indistinguishable from *"stores nothing"*), or drop it from this PR and correct #1070's description. Publishing a capability flag nothing consumes is the kind of thing that gets treated as load-bearing two releases later.

---

## Should fix

### 3. Retrieval materialises the whole archive three times over

`get_project_snapshot()` does `read_blob()` → `base64.b64encode(...)` → `jsonify(...)`. For a snapshot at the 100 MB cap that is ~100 MB of bytes, ~133 MB of base64 `str`, and Flask serialising the whole JSON document into another buffer before the response is written. Peak is comfortably north of 350 MB on a device that is often a Pi.

The base64-in-JSON choice is well argued (the agent's proxy decodes JSON or falls back to text — a binary body genuinely does not survive that trip) and I am not asking you to change the wire format. But the cap and the transport were chosen independently: 100 MB is defensible for *storage*, and indefensible for *this retrieval path*. Either lower the cap to something the retrieval path can actually serve on target hardware, or stream the base64 encoding (`Response(generator)` chunking the file) so peak memory is bounded by the chunk size rather than by the file.

There is also no rate limiting on the endpoint, so an admin — or a stuck client retrying — can repeat the allocation.

### 4. `repair_missing_admin()` is a silent privilege escalation at boot

`webserver/restapi.py`. The dead-end it fixes is real and the reasoning is sound: once any user exists, `create-user` needs an admin and `update-user` refuses a role change from a non-admin, so a device with accounts and no admin has no API path back. The refusal to guess when several accounts exist is the right call and is well tested.

My concern is the single-account case. The promotion is unconditional on *why* the account has no admin role. A device deliberately provisioned with one restricted operator account — a panel where the integrator wanted exactly that and nothing more — is silently promoted to full admin on the next boot after upgrade, and the only trace is a `logger.warning`. The docstring's justification ("a single-user device has no ambiguity about who the administrator is") is an assumption about deployments, not a property of the data.

Two ways to narrow it, in order of preference:

- Gate on evidence of the bug rather than on the symptom: the affected devices came through a build whose `role` column was nullable-with-no-default. `role IS NULL` is distinguishable from an explicit `'user'`, and `test_a_null_role_counts_as_no_admin_rather_than_crashing` shows you already handle the null case. Repairing only `NULL` fixes the devices that were actually broken and leaves a deliberately-restricted account alone.
- Failing that, make it opt-in via an env var, or emit the repair to the runtime's user-visible log rather than only to the file logger.

This one belongs in the CRA for the demand either way — it is a change to who holds administrative authority on a fielded device, applied automatically.

### 5. `ProjectSnapshotWarning` is returned and read by nobody

`handle_upload_file` now returns `ProjectSnapshotWarning` carrying strings like `"Snapshot ignored: snapshot_metadata is not valid JSON"`. Grepping both clients: neither reads it. Both build their own warnings client-side from `missingLibraries`, which is a different thing entirely — the runtime's reason for *refusing* a snapshot never reaches a user.

The consequence is the silent-failure mode this feature is least able to afford: the upload succeeds, the program runs, the user is told nothing, and the project is quietly not retrievable. Either surface it in the clients (the editor's compile log already carries this class of message) or drop the field.

### 6. `promote()` is two `os.replace` calls, not one transaction

```python
os.replace(_STAGED_BLOB, _PROMOTED_BLOB)
os.replace(_STAGED_META, _PROMOTED_META)
```

A power cut between the two leaves a promoted blob with no promoted metadata. `read_metadata()` returns `None` in that case and `read_blob()` is never consulted alone by the endpoint (it checks `record is None or blob is None`), so the device *fails safe* — it reports no stored project. Good.

But the failure is silent and permanent: the blob is orphaned on disk until the next upload's `clear()`, and a device that was storing a project now reports nothing with no log line explaining why. A one-line guard in `read_metadata()` — log at `warning` when `_PROMOTED_BLOB` exists and `_PROMOTED_META` does not — turns an invisible state into a diagnosable one. Worth it on a device you cannot attach a debugger to.

---

## Nits

- `stage()` calls `clear()` internally, and `handle_upload_file` already called `clear()` at the point of no return. Harmless (the second `rmtree` hits `FileNotFoundError` and passes) but the double-clear makes the ownership of the lifecycle harder to read than it needs to be. A comment in `stage()` noting the caller has usually already cleared would do.
- `advertised_fields()` does two filesystem stats and a JSON parse on **every** discovery probe, from inside the unauthenticated UDP responder. `PER_IP_RATE_LIMIT_SECONDS` bounds it per source and the `_last_seen` GC caps the memory, so this is not a real DoS — but a spoofed-source flood still turns each packet into disk I/O. Caching the two strings in memory and invalidating on `promote()`/`clear()` costs nothing and removes the question.
- `_clean_string` uses `str.isprintable()`, which drops tabs and newlines (correct here) and keeps non-ASCII printables (also correct — project names carry accents in this shop). Worth a one-line comment saying the non-ASCII retention is deliberate, since `isprintable()` reads like an ASCII filter at a glance.
- The import reshuffles in `plcapp_management.py` and `app.py` (`apply_vpp_plugin_conf` moved, stdlib block resorted) are unrelated tidying inside a feature PR. Not worth reverting, just noting it costs the diff some signal.

---

## What is good, and worth saying

- **The lifecycle is genuinely well reasoned.** Clearing at the point of no return rather than on arrival is the correct place, and the argument for it — *"a rejected upload leaves the previous program AND its snapshot untouched, which is the pair that is still true"* — is the right invariant stated in one sentence. `test_a_stranded_staged_snapshot_is_never_promoted_by_a_later_build` covers the exact bug the description says was a self-review catch, which is the only kind of self-review claim worth believing.
- **"An upload with no snapshot erases the stored one" is the right default and is stated as a deliberate choice**, not left implicit. `openplc-cli` and third-party clients are the reason, and a device advertising a project it is not running is the failure this avoids.
- **`normalize_metadata` drops unknown keys rather than storing them.** Everything in that record is echoed to other clients, so the shape being the device's and not the uploader's is the correct boundary, and the string/length caps are in the right place.
- **`docs/SECURITY.md` states the limitations without softening them** — not encrypted, one role check and nothing behind it, no integrity guarantee, project name readable unauthenticated. That section is a better artefact than most of what the feature does, and it makes the CRA straightforward to write.
- **Test naming reads as behaviour**, and the negative cases (`test_without_the_repair_there_is_no_way_back`, `test_retrieval_404s_while_a_snapshot_is_only_staged`) prove the properties that matter rather than the ones that are easy.

## Suggested merge order

This PR first, then orchestrator-agent #125, then the clients. The clients are inert against a runtime without these endpoints, and #125 is inert against a runtime that does not advertise `project_name`.

thiagoralves:
Fixed in f360963, with two deliberate exceptions below.

**1. Snapshot buffered before its cap (blocking).** Fixed. The part is now read with a bound (`read(cap + 1)`) and refused past it, with `MAX_CONTENT_LENGTH` set as a backstop under the whole request so the body is capped before Werkzeug parses or spools it. `content_length` is used only as a fast path, for the reason you gave — it is client-supplied and often absent on a part.

You also identified exactly why the existing test hid this: it exercised `stage()`, which proves the cap fires *after* the bytes exist. The new test asserts through a stream that refuses to serve more than the guard may ask for, so it fails against the old unbounded read rather than passing on the same error message. I checked that by reverting the fix — it fails.

Verified on hardware (Pi-class, 1845MB): a 101MB part is refused with `Snapshot ignored: archive is too large`, and device memory moved 567MB → 585MB, i.e. ordinary compile activity rather than a 101MB spike.

**3. Triple materialisation (should-fix).** Addressed without changing the wire format. The response is now streamed — base64 is encoded straight into it in 3MB chunks — so peak is bounded by the chunk rather than the file, and the bytes on the wire are identical. Verified on the device: valid JSON, valid ZIP, contents intact. The 100MB cap stays.

**5. `ProjectSnapshotWarning` read by nobody** and **2. `capabilities.projectSnapshot` read by nobody.** Both confirmed by grep across the client branches. Not yet wired — flagged to the maintainer with the recommendation to surface the warning in the clients' compile log, since a silently unretrievable project is the failure mode this feature can least afford. Tracking rather than dropping.

**4. `repair_missing_admin` (should-fix).** Deliberately unchanged, and I checked the premise rather than assuming. A no-admin state is unreachable through the API: `delete-user` is admin-only and refuses self-deletion (so the acting admin always survives), and `update-user` returns 409 `Cannot demote the last remaining admin`. The deliberately-provisioned single-operator panel you describe therefore cannot exist on a device that has ever had an admin — which every device has. The repair only ever fires for devices broken by the legacy nullable-role column, which is the population your preferred narrowing would also have targeted.

**6. Non-atomic `promote()`.** Fixed as you suggested: `read_metadata()` still reports "nothing stored" (fails safe) but now logs a warning naming the orphaned blob, so the state is diagnosable on a device you cannot attach a debugger to.

Nits: `stage()` now documents that it owns the clear rather than inheriting it; `advertised_fields()` is cached in memory and dropped by the four functions that write the store (with tests that a staged project is never advertised and that the cache follows every change — the staleness risk a cache introduces is the thing worth pinning); `_clean_string` says why non-ASCII is kept. I reverted one unrelated reformat black had made, per your note about diff signal.

Suite: 191 passing.

PR #187, opened 2026-09-01
refactor(retain): one driver contract, and the project owns the settings [NODE-94]

thiagoralves:
Runtime third of a set with **openplc-editor** and **openplc-web**. Follow-up to the NODE-94 review on editor #1034 / web #691.

Two changes, both about moving a decision to the layer that can actually make it.

## One driver interface

`plc_retain.cpp` branched on a `Backend` enum between two different call paths for one contract — `plugin_driver_retain_*` for plugins, `plc_retain_file_store_*` for the built-in. The runtime knew what kind of storage it had, in three places. It is now a single `RetainDriver` record of three function pointers, filled once at init, and nothing past that point asks which kind it got.

**The built-in file store is a driver, not a fallback branch.** Replacement needs no priority table: the file store offers itself only when `retain.conf` enables it, and the editor emits no `retain.conf` for a target whose VPP declared it owns retention — so a correctly declared device presents exactly one candidate and the existing first-wins resolution is untouched.

## The store decides what is stale

`plc_retain_load()` → `plc_retain_read()`, handed `strucpp_program_md5`. A driver finding a different program's values discards them, logs it, and reports empty. The file store writes that identity as a 32-byte header ahead of the payload, published by the same write-and-rename so the two can never disagree.

This closes a gap the layout hash cannot: **two programs that share a retained-variable layout inheriting each other's state.** It also removes any need for something outside the runtime to notice an upload — which is what makes baremetal and runtime v4 behave identically, since baremetal has no webserver to consult.

## The project owns the settings

`GET`/`PUT /api/retain-config` is gone. A device configured out of band disagreed with the project running on it, and the project is what a user can see. `apply_retain_conf` installs what the upload carries and **deletes the device's copy when nothing arrives** — that absent case is how a VPP-owned-retention target switches the built-in store off.

Consequence worth flagging: re-uploading the **same** program now keeps its retained values. The old unconditional `clear_retained()` on every upload contradicted that, so it is removed.

## Verified on hardware

An SLM-RP4, service stopped, sources synced, `sudo ./install.sh` re-run.

| Check | Result |
|---|---|
| `retain.conf` from the project | Replaced a stale REST-written config with the project's stanza |
| Cold boot (`systemctl reboot`) | Retained counter continued; control counter restarted from 0 |
| Stop / start | Same — identical behaviour to the cold-boot path |
| Same program re-uploaded | Values **kept** |
| Different program uploaded | **Storage cleared.** Same `layout 78561fad` both times, only the MD5 differing |

**One design bug hardware caught:** the editor emits `path=` for "use this device's default", and byte-copying the upload shipped the empty value to a core that treats enabled-with-no-path as a misconfiguration — retention silently off. Fixed by installing the *parsed* stanza; three regression tests cover it.

## Tests

`tests/pytest/restapi/test_retain_config.py` removed with the endpoints; `tests/pytest/plugins/test_apply_retain_conf.py` added (10 cases). **141 passed** across restapi + plugins on the device, `plc_main` compiles and links.

## Breaking

`retain_load` gains the identity arguments and `retain_clear` becomes `retain_flush`. No in-tree plugin implements them. `core/src/drivers/README.md` — the vendor-facing page — is updated.

## Merge order

This must ship **before or with** the editor half.


Gustavohsdp:
## Code review — NODE-94 follow-up (runtime)

Reviewed with the sibling PRs openplc-editor#1073 and openplc-web#723. Every line reference below was read in the file on this branch rather than inferred from the diff.

**Verdict: request changes, on one narrow logic gap plus the documentation gate.** This is the strongest piece of the three. The single driver record, the identity published with the payload, and moving the staleness decision into the store are all the right calls, and the reasoning in the comments holds up under reading.

Worth stating up front, because a page with no red marks implies otherwise: **this repository runs no CI on pull requests.** There are no checks on this PR, so the "141 passed" and "`plc_main` compiles and links" in the body are author-reported and have not been independently verified. That is not a criticism of the numbers, it is what the absence of a gate means here. The hardware runs are the real evidence in this PR, and they are good ones.

---

## 🟡 Required — the no-MD5 path leaves retain half-on, and logs a misleading warning forever

`plc_retain.cpp:181-199`:

```c
void plc_retain_read(void)
{
    if (!g_active.load() || !driver_bound()) return;
    ...
    static const uint16_t MD5_HEX_LEN = 32;
    if (!ext_strucpp_program_md5)
    {
        log_warn("Retain: the program exports no MD5 — retained variables start at their "
                 "initial values");
        return;                          // <-- g_active stays true
    }
```

The early return is correct in intent — refusing to restore without an identity is the right decision, and the comment argues it well. But it leaves `g_active == true`, and `plc_retain_save()` at `:227` is not gated on the read having happened:

```c
void plc_retain_save(void)
{
    if (!g_active.load() || !driver_bound()) return;   // still true
    const size_t n = ext_strucpp_retain_pack(g_buffer.data(), g_buffer.size());
    if (n == 0) return;
    g_driver.write(g_buffer.data(), (uint16_t)n);      // runs every scan
}
```

So saves continue every cycle. Meanwhile `g_program_md5` is assigned inside `plc_retain_file_store_load()`, which was never reached, so it stays empty. `flush_loop` snapshots that empty identity and hands it to `commit()`, which refuses at `plc_retain_file_store.cpp:139`:

```c
bool wrote = program_id.size() == PROGRAM_ID_LEN &&
             fwrite(program_id.data(), 1, PROGRAM_ID_LEN, f) == PROGRAM_ID_LEN;
if (wrote) wrote = fwrite(buf, 1, len, f) == len;
```

and falls into `:148-152`:

```c
if (!wrote)
{
    remove(tmp.c_str());
    log_warn("Retain: short write to %s — keeping the previous stored values", tmp.c_str());
    return;
}
```

Net effect: a program with no MD5 export produces a `short write` warning every `flush_seconds` — 5 by default — indefinitely, for a condition that is not a short write and that the operator can do nothing about. It also does useless pack-and-copy work on every scan for bytes that can never be committed.

The comment at `:135-138` in the file store anticipates part of this ("g_program_md5 is empty only if something committed before any load ran, which the call order rules out") — the call order does not rule it out on this path, because the load is skipped rather than failed.

### Fix

Make retain fully inert instead of half-on, so the one warning already logged is the whole story:

```c
    if (!ext_strucpp_program_md5)
    {
        /* No identity to compare against means a driver cannot tell a new
         * program from the old one, and restoring on that basis is how one
         * program inherits another's state. Refuse — and stand the store down
         * rather than leave saves running, since a blob that cannot be labelled
         * is a blob that can never be committed. */
        log_warn("Retain: the program exports no MD5 — retained variables start at their "
                 "initial values");
        g_active.store(false);
        return;
    }
```

**How much this matters:** the comment at `:186` says the symbol is resolved from the `.so` at load time and is already available here, so this is a defensive branch rather than one reached in normal operation. I am raising it as required rather than a nit because the failure mode is an unbounded stream of a warning that names the wrong cause, which is the kind of thing that costs someone an afternoon on a customer box, and the fix is one line.

---

## 🟡 Requirements Gathering and Cybersecurity Risk Assessment are missing

Neither is linked in the body, and I checked Confluence rather than assuming: `text ~ "NODE-94"` returns exactly one page, the *VPP per-device licensing* Requirements Gathering, which only mentions the key in passing. There is no NODE-94 Requirements Gathering and no NODE-94 Cybersecurity Risk Assessment.

This targets `development`, so the full bar applies, and the assessment is per repository — this one needs its own, separate from the editor's and web's. The security-relevant surface here is real and worth writing down: `GET`/`PUT /api/retain-config` removed, two IPC commands removed from `unix_socket.c`, and a settings file that now arrives inside a program upload rather than over an authenticated REST endpoint. That last one is a genuine change in how configuration reaches the device and is exactly what an assessment exists to record.

Not a reason the work should have waited. But it has to close before merge. If the position is that a follow-up inherits the parent demand's assessment, say so in the body with a link.

## 🟢 Suggestions

- `PROGRAM_ID_LEN` (`plc_retain_file_store.cpp:35`), `MD5_HEX_LEN` (`plc_retain.cpp:190`) and baremetal's `OPLC_RETAIN_PROGRAM_ID_LEN` are the same 32 declared three times, tied together only by comments. The baremetal one is across a process and toolchain boundary so a shared header is not worth it, but the two on this side could carry a `static_assert` against each other.
- `plc_retain_init()` writes `g_driver` (four pointers) while `plc_retain_save()` reads it from the scan thread. The publish is safe as written — `g_active.store(false)` before the mutation and `store(true)` after, both `seq_cst`, with every reader checking `g_active` first — so this is correct today. It is worth a line of comment saying that `g_active` is the publication barrier for `g_driver`, because the next person to add an early return that skips the store, or to relax the memory order, will not derive it from the code alone. Low confidence that this ever bites; raising it as a documentation point, not a defect.

---

## ✅ What is good here

- **One driver record, and nothing downstream asks what kind of store it got.** Replacing the `Backend` enum and its three branch sites with a four-field struct filled once at init is a real simplification, and the claim in the comment — that a third kind of store means filling this in from somewhere new and changing nothing else — is actually true of the code as written.
- **Rank, not arbitration, and the reason it works.** The file store offers itself only when `retain.conf` enables it, and the editor emits no `retain.conf` for a target whose VPP owns retention, so a correctly declared device presents exactly one candidate. That is a much better property than a priority table, and it is the kind of invariant that is easy to state and hard to arrange.
- **Identity and payload published as one unit.** `[32 bytes of md5 hex][blob]` through the same write-and-rename means a file whose header says program A is guaranteed to contain program A's values; a sidecar would have had a window where it does not. Holding the identity from `load()` and committing it on the next `save()`, so a read never mutates storage, is the right sequencing and the comment at `:281-290` explains why.
- **`fsync` on the directory after the rename** (`plc_retain_file_store.cpp:162-175`). Correct, load-bearing on ext4, and the comment names the exact failure it prevents — the store coming back with the previous blob even though the new one was written. This is only ever added by someone who has lost data to it.
- **A torn or pre-header file fails the identity check and is discarded** rather than being read as a blob. The right outcome for bytes whose owner cannot be established, and it means the format change needs no migration.
- **`plc_retain_flush()` placed between `pthread_join` and `plugin_driver_stop`** (`plc_state_manager.cpp:1144`), with the comment explaining why both bounds matter — after the join so no scan is mid-save, before the stop so a plugin-backed store is still alive to answer. Easy to get wrong in a way no test would catch.
- **"Save is the durability path; flush is only a hint"** in `plugin_driver.h`. Stating that a plugin committing solely in `flush()` loses everything in exactly the case retention exists for is the single most useful sentence in the vendor-facing contract, and it was not there before.
- **Deleting FC 0x4C on evidence.** No caller anywhere, could not have worked if it had one because the per-scan save overwrote a post-upload clear within a cycle, and it skipped CRC while being the one command in the set that destroys user data. Removing it and making clearing a consequence of the identity check is strictly better than fixing it.
- **The hardware run has a negative control.** Same retain layout `78561fad` both times with only the MD5 differing is the right experiment for the claim, because it isolates the identity gate from the layout hash. Reporting the design bug the hardware caught — the editor emitting `path=` and byte-copying shipping the empty value into a core that treats enabled-with-no-path as a misconfiguration — with three regression tests, is the part of this PR I would point at as the model.
- **`tests/pytest/plugins/test_apply_retain_conf.py` is in `tests/pytest/plugins/`**, which `scripts/run-pytest.sh` already invokes as its second pass, so the new suite actually runs.
- **`core/src/drivers/README.md` updated in the same PR** and the breaking change declared. The vendor-facing page changing with the contract is the thing that most often lags.

---

## Previous review requests (NODE-94) — item by item

| Request | Status | Evidence |
|---|---|---|
| **A1** — `MB_FC_RETAIN_RESET` / FC 0x4C had no caller and could not have worked | ✅ Closed | Function code deleted; `RETAIN:CLEAR` and `RETAIN:STATUS` removed from `unix_socket.c` |
| **A2** — `capacity()` and `clear()` do not belong in the driver contract | ✅ Closed | Three-function contract; `capacity()` replaced by `write()` returning `TOO_LARGE`, `clear()` by the identity check in `read()` |

No regressions found. The one required item is a defensive branch that was not carried all the way through, not something the refactor broke.

---

## To unblock

1. `g_active.store(false)` on the no-MD5 path in `plc_retain.cpp:198`
2. Link the Requirements Gathering and the Cybersecurity Risk Assessment for this repository, or state in the body why this follow-up inherits the parent's
3. Land this before or with editor#1073 and web#723, as the body already says

Since this repository has no PR CI, I did not run the C build, the Ceedling suite or `scripts/run-pytest.sh` as part of this review — the reviewer can ask for that separately and it is worth more here than in the other two repos. Happy to re-review once pushed.

webserver/plcapp_management.py
```diff
@@ -425,6 +433,95 @@ def apply_vpp_plugin_conf(generated_dir: str = "core/generated") -> None:
...
+    # validated here is exactly what the core will read back.
+    cfg = read_retain_conf_file(uploaded_conf)
+
+    try:
+        if cfg["enabled"]:
+            # Only meaningful when the store is on. A disabled stanza carrying a
+            # path that happens not to exist is not an error worth refusing an
+            # upload over -- nothing will be written there.
+            validate_retain_path(cfg["path"])
+        validate_flush_seconds(cfg["flushSeconds"])
```
dcoutinho1328: `flushSeconds` is validated even when the store is disabled, so an out-of-range value on a *disabled* stanza refuses the whole upload — not just the unused field.

```python
if cfg["enabled"]:
    # Only meaningful when the store is on...
    validate_retain_path(cfg["path"])
validate_flush_seconds(cfg["flushSeconds"])
```

The same reasoning that exempts `path` when disabled ("nothing will be written there") applies just as well to `flushSeconds` — it's dead while `enabled=0`. Reproduced against this branch:

```python
plcapp_management.RETAIN_CONF_PATH = dest   # dest pre-populated with a previous project's config
# upload's retain.conf: enabled=0, path=/tmp/whatever.bin, flush_seconds=<MAX_FLUSH_SECONDS + 1>
plcapp_management.apply_retain_conf(upload)
os.path.exists(dest)
# -> False
```

Result: a project with persistent storage turned off but a stale/out-of-range `flushSeconds` (e.g. after `MAX_FLUSH_SECONDS` is tightened in a later release, or a hand-edited project file) has its entire `retain.conf` refused on every upload, and any previously-installed config removed — for a value that is never read while disabled.

Suggested fix: only validate `flushSeconds` inside the `if cfg["enabled"]:` block, same as `path`.

core/src/plc_app/plc_retain_file_store.cpp
```diff
@@ -228,16 +281,48 @@ int plc_retain_file_store_save(const uint8_t *blob, uint16_t len)
...
+    const size_t got_id = fread(stored_id, 1, PROGRAM_ID_LEN, f);
+    if (got_id != PROGRAM_ID_LEN)
+    {
+        /* Shorter than the header: a torn write, or a file written before the
+         * header existed. Either way its owner cannot be established. */
+        fclose(f);
+        discard_stored();
+        log_info("Retain: stored values could not be attributed to a program — storage cleared");
+        return 0;
+    }
```
dcoutinho1328, line 318: No automated test exercises the identity-mismatch path this function implements.

`tests/pytest/plugins/test_apply_retain_conf.py` (new in this PR) covers the webserver's install/validate logic thoroughly, but nothing covers the core behavior change here: the `[32-byte md5][blob]` header, the discard-on-mismatch at this `memcmp`, the torn/pre-header fallback a few lines up, or `plc_retain_flush()`'s placement between `pthread_join` and `plugin_driver_stop` in `plc_state_manager.cpp`. The PR body's hardware-test table is the only evidence for this path.

The repo already has a host-runnable harness suited to exactly this (`tests/lifecycle/`, `fakevpp_plugin.c` + `suite.py`), and it wasn't touched — no plugin-store or file-store retain scenario in it. Worth a lifecycle case (or a small standalone host test) that: writes a stored blob under one program's identity, loads with a different MD5, and asserts the discard + empty-result behavior — the same "different program uploaded → storage cleared" case verified by hand on hardware in the PR body, made repeatable without it.

dcoutinho1328:
Cross-checking Jira: this PR is filed under **NODE-94**, but that ticket's own scope explicitly excludes what this PR does: "Out: the runtime-side persistence, crash-safe store, and survival across updates (RTOP, FR9/NFR3)." NODE-94's acceptance criterion is editor-UI only ("A user can mark variables as retain and set their default values in the editor... saved in the project"), which this PR doesn't touch.

The ticket that matches this work exists: **RTOP-210** — "Weidmüller u-OS: retain-variable persistence, crash-safe store, and survival across updates" (FR9/NFR3) — but it's still in Backlog, with nothing linked to this PR.

This is likely why no NODE-94 Requirements Gathering or Cybersecurity Risk Assessment exists yet: the work is filed under a ticket whose scope excludes it. Worth re-linking this PR to RTOP-210 (and moving it out of Backlog) rather than writing a NODE-94 RG/CRA for work NODE-94 explicitly scopes out.

webserver/plcapp_management.py
thiagoralves: Fixed in 60d7da6 — you were right, and the `MAX_FLUSH_SECONDS`-tightening case is the one that convinced me.

`validate_flush_seconds` is now inside the `if cfg["enabled"]:` block alongside `validate_retain_path`, for the reason you give: neither is read while `enabled=0`, and refusing deletes the device's existing config over a field nothing consults.

Test added — `test_a_disabled_stanza_is_installed_without_validating_its_flush_period` — which is the repro you posted, asserting the stanza *is* installed rather than the config vanishing. 11 passing.

core/src/plc_app/plc_retain_file_store.cpp
thiagoralves: Fixed in 60d7da6 — fair hit, the hardware table was doing work a test should do.

I went with the standalone host test rather than a lifecycle case. `tests/lifecycle/` could reach this, but it boots a real `plc_main` against a compiled program and needs Linux, Docker and an editor payload — more machinery than file-header logic warrants, and it wouldn't run on a developer's machine. Ceedling can't reach it either (configured for C; this store uses `std::thread`/`std::mutex`).

So `tests/host/` is new: plain executables, one command, `./tests/host/run.sh`. Seven cases covering what you named —

- the `[32-byte md5][payload]` layout;
- the same program getting its values back;
- **a different program's values discarded, not inherited** — the case a layout hash cannot catch, and it also asserts the next commit carries the *new* identity, so retention actually works for the new program;
- a file too short to carry a header treated as unattributable;
- an empty store reported as empty rather than failing;
- a wrong-length identity refused;
- an unchanged blob not re-committed (inode unchanged), so the dedupe can't be quietly dropped.

Two of them assert the operator is *told* storage was cleared, since a silent discard of retained values is what this design would get blamed for.

Writing it caught a documentation error of mine: this store is `[32-byte md5][payload]` with the length implicit in the file size — the explicit 2-byte length is baremetal's *flash* driver, whose region is fixed-size and whose erased bytes read as 0xFF. My first test asserted the wrong format and the code was right. Both formats now stated correctly.

`plc_retain_flush()`'s placement between `pthread_join` and `plugin_driver_stop` is still uncovered — that one genuinely needs the lifecycle harness, and I'd rather add it there deliberately than fake it here.

thiagoralves:
## Review addressed — pushed in 60d7da6

### Fixed

**R1 — the no-MD5 path left retain half-on.** Correct in every detail, including the chain through to the `short write` warning. `g_active.store(false)` added, so the one warning already logged is the whole story instead of the opening of an unbounded stream naming the wrong cause. Your point that the file store's comment claiming "the call order rules out" an empty identity was *false on this path* is the part that made it clearly a bug rather than a tidy-up — that comment is true again now.

**R2 — `flushSeconds` validated while disabled.** Fixed with a test that is @dcoutinho1328's repro; replied inline.

**R3 — nothing covered the store's identity handling.** `tests/host/` added, seven cases, `./tests/host/run.sh`. Replied inline with the reasoning for a standalone host test over a lifecycle case, and with one thing writing it caught: **my own format documentation was wrong.** This store is `[32-byte md5][payload]` with the length implicit in the file size; the explicit 2-byte length is baremetal's flash driver, whose region is fixed-size. My first test asserted the wrong format and the code was right.

**R5 — both suggestions taken.** The 32 is now declared once as `PLC_RETAIN_PROGRAM_ID_LEN` in the contract header both sides include, with a `static_assert` where the file store derives its on-disk header from it — which removes the duplication rather than just asserting over it. Baremetal's copy stays separate on purpose, for the process/toolchain-boundary reason you gave. And `g_active`'s role as the publication barrier for `g_driver` is now written down; you're right that it's correct today and not derivable from the code by whoever next adds an early return.

### Still open, and above me

**The Requirements Gathering and Cybersecurity Risk Assessment.** @dcoutinho1328's cross-check is the useful frame here, and I verified it: **RTOP-210** exists — *"Weidmüller u-OS: retain-variable persistence, crash-safe store, and survival across updates"*, FR9/NFR3, status **Backlog**, scope *"In: the runtime-side persistence mechanism and crash-safe store"*. NODE-94's scope says the opposite: *"Out: the runtime-side persistence, crash-safe store, and survival across updates (RTOP, FR9/NFR3)"*.

So this PR is filed under a ticket that explicitly scopes it out, which likely explains why no NODE-94 RG/CRA exists.

One caveat on re-filing straight to RTOP-210: that ticket is specifically u-OS / NV-RAM ("Do not assume the storage API"), while this PR is generic runtime-v4 plus bare-metal. Neither ticket is a clean fit, so I'd rather flag it than quietly move it — this needs someone to decide whether RTOP-210 widens, or a new ticket covers the generic mechanism with RTOP-210 as the u-OS consumer.

### On the no-CI point

Accurate, and worth having said. This repo has no PR workflow, so the numbers in the body are author-reported. What I can offer is that they are now reproducible without hardware: `./tests/host/run.sh` and `pytest tests/pytest/plugins/test_apply_retain_conf.py` both run on a laptop. **11 pytest + 7 host cases passing**, `plc_main` builds and links on the SLM-RP4.

marconetsf (approved):

PR #192, opened 2026-09-04
fix(install): show the download, and pin the version latest resolved to

thiagoralves:
Installer-only follow-up to v4.2.2, from testing the published one-liner
(`curl -fsSL https://runtime.getedge.me | sudo bash`) on a real device.

## Three fixes

**The download was silent.** `docker pull` ran with output suppressed, so
the installer sat on a single `Pulling` line for as long as the image took
to arrive. On a slow link that reads as a hang, and it got killed. Docker's
own per-layer progress is shown now. The wait that follows also reports the
bootloader's state, its reason, and elapsed seconds -- rewritten in place on
a terminal, appended once per change when it is not one.

**The long download happened in the device's no-PLC window.** Both images
are now pulled before anything is disturbed, runtime first. Nothing
destructive before the images are on disk was already the intent, but the
runtime image was left for the bootloader to fetch *after* the native
runtime had been stood down.

**`latest` was what got recorded.** Installing the newest release is the
right default; writing `latest` into the device spec is not -- the device
would report its version as `latest` and follow the tag on every reconcile,
moving to a new runtime because someone published one rather than because
anyone chose it here. The concrete version is now read out of the image's
own config (`RUNTIME_VERSION`, baked in by the release build), so it needs
nothing beyond the registry and is by construction exactly what `latest`
resolved to. If it can't be read, the installer warns and keeps the moving
tag rather than guessing.

## Verified

In a privileged real-systemd container with an active
`openplc-runtime.service` to displace, against the live registry:

- spec records `"version": "v4.2.2"`, and the bootloader agrees
  (`{"runtimeVersion":"v4.2.2","state":"healthy"}`)
- the pull prints its progress, digest and status
- `bash -n` clean, `shellcheck -S warning` clean

### DIFF
--- a/scripts/install-docker.sh
+++ b/scripts/install-docker.sh
@@ -3,24 +3,19 @@
 #
 # Two ways in, one script:
 #
-#   curl -fsSL https://runtime.getedge.me | sudo bash      no checkout, no toolchain
+#   curl -fsSL https://runtime.getedge.me | sudo bash      no checkout needed
 #   sudo ./install.sh                                      from a clone; execs this
 #
-# It installs no build toolchain and compiles nothing: it ensures a container
-# engine, writes the bootloader's spec, and starts the bootloader. The
-# bootloader then pulls the runtime image and brings it up.
+# Compiles nothing: it ensures a container engine, writes the bootloader's
+# spec, and starts the bootloader, which pulls the runtime image and brings it
+# up. Docker is the only dependency this path adds, and nothing of ours goes
+# into systemd -- Docker's restart policy starts the bootloader at boot.
 #
-# Docker is the ONLY dependency this path adds. Nothing of ours goes into
-# systemd -- Docker's own restart policy starts the bootloader at boot, and the
-# bootloader starts the runtime. That is deliberate: the fewer moving parts
-# between power-on and a running PLC, the fewer ways it fails.
+# --native keeps the source build, for MSYS2 and targets that cannot host a
+# container engine. It needs the repository on disk, so the piped one-liner
+# cannot reach it.
 #
-# `sudo ./install.sh --native` keeps the source build, for MSYS2 and for
-# targets that cannot host a container engine. That path needs the repository
-# on disk, so it is not reachable from the piped one-liner.
-#
-# `--uninstall` removes everything this script created and puts back whatever
-# it displaced, so a device ends up as it started.
+# --uninstall removes what this script created and puts back what it displaced.
 set -euo pipefail
 
 RED='\033[0;31m'; GREEN='\033[0;32m'; YELLOW='\033[1;33m'; BLUE='\033[0;34m'; NC='\033[0m'
@@ -38,13 +33,10 @@ BOOTLOADER_VERSION="${BOOTLOADER_VERSION:-latest}"
 
 # Two directories, deliberately separate.
 #
-# The runtime's holds everything a version change must preserve: .env,
-# restapi.db, retain.bin, vpp/ licences and the stored project.
-#
-# The bootloader's holds the container spec, including this board's device
-# mounts. It is separate because "erase all data" wipes the runtime's
-# directory, and a board that came back from a data wipe with no SPI would be
-# a miserable failure mode.
+# The runtime's holds what a version change must preserve: .env, restapi.db,
+# retain.bin, vpp/ licences, the stored project. The bootloader's holds the
+# container spec, including this board's device mounts -- separate because
+# "erase all data" wipes the runtime's directory.
 RUNTIME_DATA_DIR="${RUNTIME_DATA_DIR:-/var/lib/openplc-runtime}"
 BOOTLOADER_STATE_DIR="${BOOTLOADER_STATE_DIR:-/var/lib/openplc-bootloader}"
 
@@ -61,15 +53,10 @@ declare -a EXTRA_ENV=()
 # which accumulates across installs and is what --uninstall reads.
 declare -a DISPLACED_THIS_RUN=()
 
-# Where this script lives, for the self-elevation path below. A piped run has
-# no file to re-exec, so it fetches a fresh copy rather than trying to re-read
-# a stdin bash has already consumed.
-# Where the piped one-liner fetches this script.
-#
-# runtime.getedge.me serves this file verbatim from the release branch. Until
-# that host is stood up the raw GitHub URL below works identically, and
-# OPENPLC_INSTALLER_URL overrides both -- which is also how a fork or an
-# internal mirror points it somewhere else.
+# Where this script lives, for the self-elevation path below.
+# Where the piped one-liner fetches this script. runtime.getedge.me serves it
+# verbatim from the release branch; the raw GitHub URL works identically, and
+# OPENPLC_INSTALLER_URL overrides both, for a fork or an internal mirror.
 INSTALLER_URL="${OPENPLC_INSTALLER_URL:-https://runtime.getedge.me}"
 INSTALLER_URL_FALLBACK="https://raw.githubusercontent.com/Autonomy-Logic/openplc-runtime/main/scripts/install-docker.sh"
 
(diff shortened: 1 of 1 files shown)


scripts/install-docker.sh
```diff
@@ -507,10 +507,41 @@ resolve_runtime_version() {
...
+    [ "$RUNTIME_VERSION" = latest ] || return 0
+
+    local image="$RUNTIME_REPOSITORY:latest" baked
+    baked="$(docker image inspect --format \
+        '{{range .Config.Env}}{{if eq (index (split . "=") 0) "RUNTIME_VERSION"}}{{index (split . "=") 1}}{{end}}{{end}}' \
+        "$image" 2>/dev/null || true)"
+
+    if printf '%s' "$baked" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+'; then
+        log_info "'latest' is $baked; recording that rather than the moving tag"
+        RUNTIME_VERSION="$baked"
```
thiagoralves, line 507: **Pinning the spec to a tag that is not on disk breaks the local-copy path.**

`pull_image` deliberately accepts "the registry is unreachable but a copy is already here" (air-gapped, `docker load`, or a device installed by an earlier run that already has `repo:latest`). In that case only the `:latest` **tag** exists locally. This function then reads `v4.2.2` out of that image and rewrites `RUNTIME_VERSION`, so `write_spec` records `"version": "v4.2.2"` — a tag the daemon has never seen.

What happens next, with no network:

1. `stop_legacy_runtimes` stands the native runtime down.
2. The bootloader reconciles. `containerIsStale` compares `inspect.Config.Image != desired` (`repo:latest` vs `repo:v4.2.2`) → stale → `recreate` **stops the working runtime container**.
3. `create` → `ensureImage("repo:v4.2.2")` → `InspectImage` not found → `PullImage` → fails → recovery.

`wait_for_runtime` returns 0 on `recovery`, so the `EXIT` trap is cleared and no rollback runs: a device that had a working PLC ends the install with none. That is exactly the failure mode the pull-before-anything-destructive ordering was added to prevent.

Even with a working link this costs a registry round-trip inside the no-PLC window: the installer pulls `:latest`, the bootloader then has to resolve `:v4.2.2` (cheap, layers are shared, but it still needs the registry). If GHCR rate-limits or the link drops between the two, same outcome.

Suggested fix: tag what was actually pulled, so the pinned ref resolves locally regardless.

```sh
if printf '%s' "$baked" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+'; then
    log_info "'latest' is $baked; recording that rather than the moving tag"
    RUNTIME_VERSION="$baked"
    # The spec now names a tag the daemon has only under ':latest'. Give the
    # image that name too, so the bootloader does not have to go back to the
    # registry in the window where the device has no PLC -- and so an offline
    # or side-loaded install resolves at all.
    docker tag "$image" "$RUNTIME_REPOSITORY:$RUNTIME_VERSION" 2>/dev/null || true
    return 0
fi
```

(Alternatively, only pin when the `docker pull` itself succeeded — but tagging is strictly better, since it also removes the round-trip in the online case.)

scripts/install-docker.sh
```diff
@@ -618,29 +661,58 @@ start_bootloader() {
...
+
+    local waited=0 state="" reason="" last_reported=""
     while [ "$waited" -lt 900 ]; do
-        state="$(curl -sk "https://127.0.0.1:$BOOTLOADER_PORT/api/bootloader/capabilities" \
-            2>/dev/null | sed -n 's/.*"state":"\([a-z]*\)".*/\1/p')"
+        local caps
+        caps="$(curl -sk --max-time 5 \
+            "https://127.0.0.1:$BOOTLOADER_PORT/api/bootloader/capabilities" 2>/dev/null || true)"
+        state="$(printf '%s' "$caps" | sed -n 's/.*"state":"\([a-z]*\)".*/\1/p')"
+        reason="$(printf '%s' "$caps" | sed -n 's/.*"reason":"\([^"]*\)".*/\1/p')"
```
thiagoralves: **`reason` is never present in this payload, so this is dead code.**

`GET /api/bootloader/capabilities` is unauthenticated and deliberately minimal — `bootloader/internal/api/server.go:367` returns only `service`, `bootloaderVersion`, `runtimeVersion`, `state` and `recovery`. `reason` is on the *authenticated* `GET /api/bootloader/status` (`server.go:467`) only.

So `$reason` is always empty: the `${reason:+ -- $reason}` suffix on the progress line never renders, and — the part that matters — `[ -n "$reason" ] && log_warning "  $reason"` on the `recovery` branch never fires. The one place an operator most needs the explanation ("downloading …", "runtime exited with 1", the supervisor's own recovery reason) still prints nothing, despite the PR description saying it does.

Either add `"reason": status.Reason` to `handleCapabilities` (it is already exposed to anyone who can reach the port via `state`/`recovery`, so it leaks nothing new beyond the prose itself — worth a deliberate call), or drop the `reason` extraction here rather than leave a diagnostic that cannot fire.

scripts/install-docker.sh
```diff
@@ -618,29 +661,58 @@ start_bootloader() {
...
+    # the bootloader's own state and the elapsed time go to the terminal as
+    # they change.
+    log_info "Starting the runtime and waiting for it to report healthy"
+
+    local waited=0 state="" reason="" last_reported=""
     while [ "$waited" -lt 900 ]; do
-        state="$(curl -sk "https://127.0.0.1:$BOOTLOADER_PORT/api/bootloader/capabilities" \
-            2>/dev/null | sed -n 's/.*"state":"\([a-z]*\)".*/\1/p')"
+        local caps
+        caps="$(curl -sk --max-time 5 \
```
thiagoralves, line 637: Minor: `waited` is no longer wall-clock now that the request itself can block.

`--max-time 5` is a good addition, but each iteration can now take up to ~10s (5s curl + 5s sleep) while `waited` still advances by 5. If the bootloader port accepts a connection but stalls the TLS handshake or the response, the progress line under-reports elapsed time by up to 2x and the loop's 900 bound becomes ~30 minutes rather than 15 — and the final message then claims "has not reported healthy after 900s" when twice that has passed.

Using the clock instead keeps both honest:

```sh
local started; started=$(date +%s)
while waited=$(( $(date +%s) - started )); [ "$waited" -lt 900 ]; do
```

thiagoralves (commented):
Review of the installer changes. Three findings, one of which is a real regression for the offline / local-copy path that this same function is documented to support.

scripts/install-docker.sh
thiagoralves: Valid, and reproduced. Fixed in 3da0932.

Blackholed ghcr.io in a real-systemd container, left only `repo:latest` on disk, and ran the installer: recovery state, no runtime container at all. Exactly the chain you described.

Took the tagging fix rather than the pull-succeeded condition, for the reason you gave — it removes the round-trip online too. One change from your snippet: the tag is not best-effort. `docker tag` failing would leave the spec naming a tag the daemon still lacks, which is the bug again, so the version is recorded only if the tag succeeds and otherwise the moving tag is kept with a warning.

Same test after the fix: spec `v4.2.2`, container `v4.2.2` healthy, registry still blackholed.

scripts/install-docker.sh
thiagoralves: Valid — `handleCapabilities` (server.go:371) returns no `reason`, so the extraction was always empty and the recovery branch printed nothing where the PR claimed an explanation.

Dropped the extraction rather than widening the unauthenticated payload. The reason describes internal failure states, and the installer runs as root on the device, where `docker logs` already has it without exposing anything to whoever can reach the port. The recovery branch now names that log.

scripts/install-docker.sh
thiagoralves: Valid. Fixed — it reads the clock now.

Measured it against a port that accepts and then stalls: counted arithmetic claimed 20s after 25s of real time; measured reports 20s at 20s. The 900s bound is now a real 900s.

PR #193, opened 2026-09-06
Fix OPC UA DataValue StatusCode compatibility

genishernandez:
## Description

Fixes OPC UA DataValue construction compatibility between asyncua versions.

asyncua 1.x uses the `StatusCode_` field, while newer versions use `StatusCode`. The synchronization layer now handles both APIs without changing the configured dependency version.

This fixes OPC UA read and subscription operations failing with:

TypeError: DataValue.__init__() got an unexpected keyword argument 'StatusCode_'

## Changes

- Added a compatibility helper for `ua.DataValue`.
- Supports both `StatusCode` and `StatusCode_`.
- Updated read callbacks and subscription push DataValue creation.

Target branch: development

### DIFF
--- a/core/src/drivers/plugins/python/opcua/synchronization.py
+++ b/core/src/drivers/plugins/python/opcua/synchronization.py
@@ -39,6 +39,17 @@ from typing import Any, Callable, Dict, Optional, Set, Tuple
 
 from asyncua import Server, ua
 
+# asyncua compatibility between 1.x and 2.x
+def _make_data_value(**kwargs):
+    try:
+        return ua.DataValue(**kwargs)
+    except TypeError as exc:
+        if "unexpected keyword argument 'StatusCode'" not in str(exc):
+            raise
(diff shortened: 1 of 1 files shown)


genishernandez:
The issue was detected while testing the OPC UA functionality of OpenPLC Runtime in an Ubuntu Linux virtual machine.

During the test, OPC UA read and subscription operations failed with the following error:

`TypeError: DataValue.__init__() got an unexpected keyword argument 'StatusCode_'`

We reproduced the issue in the Ubuntu environment and traced it to the `asyncua` `DataValue` API. The installed asyncua version expects `StatusCode`, while the synchronization layer was constructing `DataValue` using `StatusCode_`.

This led us to implement a compatibility helper that supports both `StatusCode` and `StatusCode_`, allowing the OPC UA synchronization layer to work with both asyncua API variants without changing the project's configured dependency version.

The fix was then tested against the affected OPC UA read and subscription paths.

PR #195, opened 2026-09-09
feat(image): allocate the I/O image on program load instead of a fixed BUFFER_SIZE

JulioSergioFS:
## Summary

`BUFFER_SIZE` is gone. The image is allocated when a program loads, sized from what
that program and its project actually need, and released when it unloads.

It was 1024 per table, compiled in, identical for every program that ever ran on the
device. That is wrong in both directions: a project needing more could not have it —
the reporter of openplc-editor#296 has a board with three times the memory and the
same ceiling — and a project needing less paid for the rest anyway, out of the memory
its own program wanted.

Six commits, each one standing on its own:

1. **One image symbol, and a tripwire.** The fourteen tables become one
   `image_tables_t g_image`, and `plugin_driver.c`'s hand-written `extern` block for
   the same fourteen is deleted — redundant while the shapes agreed, two incompatible
   declarations in different TUs the moment they stopped, which C does not diagnose
   across translation units. The `sizeof` on those tables is confined to one function
   and pinned by `static_assert`s.
2. **`image.conf` and its install-time gate.** `apply_image_conf` beside
   `apply_retain_conf`, validating at install with a line in the build log, for the
   reason that function's own docstring already gives.
3. **The floor derived from the loaded program.** `max(byte_index)+1` per table from
   the `.so`'s `locatedVars[]`, adopted as `max(configured, derived)`.
4. **The allocation**, between `plugin_manager_load` and `plugin_driver_init`.
5. **The forced-slot map** sized from the image instead of its own 1024.
6. **The Python plugin limits**, which are what make any of this visible to a user.

## Ticket

RTOP-284 — https://autonomylogic.atlassian.net/browse/RTOP-284
Requirements Gathering, approved v1.2: https://autonomylogic.atlassian.net/wiki/spaces/CD/pages/282886145
Implementation Plan, with the phase breakdown and every decision recorded:
https://autonomylogic.atlassian.net/wiki/spaces/CD/pages/288292865
Editor side, in review: DOPE-615 — Autonomy-Logic/openplc-editor#1093 and
Autonomy-Logic/openplc-web#742. That task emits the sizes this one reads.

## Decisions worth a reviewer's attention

**One size for all fourteen tables.** `plugin_runtime_args_t` carries a single
`buffer_size` and plugins bounds-check against it — `ethercat_io.c` refuses a
`byte_index` at or above it, `s7comm` derives every clamp from it. One number describes
fourteen tables only while they are the same size. Give each its own and no value of
that field is right: the minimum makes every plugin refuse everything the moment one
table is empty (a project with `%QW4096` and no `%IX` has a floor of zero), and the
maximum lets a plugin write past the end of the smaller ones — the exact overflow this
work prevents. Per-table sizes need a field per table, which breaks CON06 and
invalidates pre-compiled plugins. So the image is square, at the largest count any
table needs: ~460 KB of pointers for a 4096-word program on 64-bit Linux, against
breaking every shipped plugin. `image.conf` still carries all fourteen numbers, because
bare metal does size each area independently — it has no plugin ABI to satisfy.

**Placement is the requirement, not the mechanism.** The allocation sits after
`plugin_manager_load`, because the floor comes from walking the loaded `.so` and there
is no `.so` before it, and before `plugin_driver_init`, because that is where the base
pointers and `buffer_size` are copied into the runtime args and both native plugins
cache that struct **by value** inside `init()`. Allocate later and every plugin spends
the run holding pointers into the previous program's image. Nothing enforced that
ordering, so building the runtime args now refuses outright when the capacity is zero,
before the pointers are copied — not an `assert()`, which vanishes under NDEBUG. The
failure it prevents is not a crash: plugins would hold null tables and a `buffer_size`
of zero, which every bounds check reads as "refuse every index", presenting as I/O that
silently does nothing.

**RSK05, the open question in the requirements, is answered from the code.** It asked
whether the Modbus server should fail, warn or clamp when the image is smaller than the
configured exposure. It is a false choice. Each data block declares itself as wide as
its counts, and pymodbus's own `validate()` answers exception 02 (Illegal Data Address)
past that. A block wider than the image lets the gap pass validate, fail the buffer
read, and answer **zero** — a plausible wrong value a client cannot tell from a real
one. Clamped to the image the two agree and the protocol reports the truth by itself,
so clamping *is* reporting once it stops lying about what exists. Failing would lose
the Modbus server entirely over a missing file, and the cause is usually a version
mismatch rather than a user error. A startup warning names what shrank, for the person
who configured it and is not the client.

**A case nobody had noticed, and it is the common one.** `generateModbusSlaveConfig`
materialises its defaults (1024 registers, 8192 coils) into `modbus_slave.json` even
when the project never opened the Modbus screen — so a project with eight `%QW` would
declare 1024 registers over an image of eight. Clamping to the image fixes it with no
editor change.

## How it was tested

- Full image build from this branch (`docker build`, which runs `install.sh --native`):
  compiles and links the whole runtime with the project's own CMake flags.
  `Built target plc_main`, **no warnings in any changed file**.
- **The runtime boots and every plugin initialises against the new allocation.**
  `./build/plc_main --print-logs` in the container logs
  `[image_tables] image allocated: 1 elements per table` — the boot minimum, allocated
  before `plugin_driver_init` — and all five plugins report PASS on `init`,
  `start_loop`, `stop_loop` and `cleanup`. That covers the boot path, the promise that
  a plugin never receives a null base pointer or a zero size, and the fact that the new
  ordering guard does not fire on the correct path.
- `bash scripts/run-pytest.sh` territory: 202 webserver tests and 24 Python-plugin
  tests. 18 new tests for `apply_image_conf` (present/absent contract, install-time
  refusal, the ABI ceiling, the parser) and 6 for the cross-implementation contract.
- `-fsyntax-only` under `-Wall -Wextra` on every changed translation unit, including
  the journal in **both** build variants (lock-free and the mutex fallback — each has
  its own `journal_init`/`journal_cleanup`, and fixing one would work only on machines
  with lock-free atomics).
- The Phase 1 `static_assert`s were verified to fire, with the intended message, by
  swapping a table to a pointer in a sandbox — and then fired for real when the types
  changed in commit 3, naming the function to follow.

**A guard for a contract that had none.** `image.conf` is written in three places and
read in a fourth: the editor emits it, the webserver validates and installs it, the
core parses it, and `image_tables.h` declares the tables it names. A key added on one
side and forgotten on another fails nothing — the core never sees that table's size,
falls back to the derived floor, and the image is quietly smaller than the project
asked for. `tests/pytest/plugins/test_image_conf_contract.py` checks the enum, the key
array and the struct fields against the Python list **in order**, by reading the C
sources as text. That is deliberate: pytest is the only suite CI runs in this
repository, so it is the only guard that will actually fire. Verified it catches a
reordering.

## Checklist

- [x] pytest passes (202 webserver + 24 plugin; the failures in
      `tests/pytest/modbus_master`, `tests/pytest/plugins/opcua` and
      `test_openplc_input_registers_datablock.py` are identical on a clean
      `development` checkout)
- [x] `pre-commit run` clean on the changed files
- [x] Docs — the reasoning lives in the headers it belongs to, and in the linked
      Implementation Plan
- [x] Follows `docs/pr-reviews/PR_REVIEW_CHECKLIST.md`

## Two things found on the way, neither introduced here

- `project.yml`'s `-DBUFFER_SIZE=128` never took effect where `image_tables.h` was
  included: the `#define` was unconditional and overrode the command line. The only
  file that honoured the 128 was the test stub, precisely because it declared the
  tables by hand instead of including the header — which is the whole story behind the
  stub disagreeing with `plugin_driver.c`. Both are fixed here.
- **`pre-commit run --all-files` reformats around 200 files in this repository**, because
  `black` and `ruff format` disagree and undo each other. Worth fixing separately; until
  then, run the hooks on your own files.


marconetsf:
**Review — RTOP-284, allocate the I/O image on program load**

Reviewed alongside the editor half (openplc-editor#1093 / openplc-web#742). The cross-repo contract holds: `kImageTableKeys` and `image_table_id_t` carry the same fourteen names in the same order as the editor's `TABLES` array, the BOOL tables are counted in bytes on both sides, the parser accepts the `#` header and treats zero as a real answer, and `image.conf` follows the same path convention as `retain.conf`. The ctypes mirror in `plugin_runtime_args.py` was checked field-for-field against `plugin_runtime_args_t` (`plugin_types.h:189-275`) and matches in order, type and count — nothing in CI checks that, so it was the biggest single risk here and it is clean. `image_tables_fill_null_pointers` runs during bootstrap before `set_realtime_priority()`, so it is not a real-time violation, and no hand-written `extern` of the fourteen tables survives anywhere.

Four blockers below, then the rest.

---

**BLOCKER 1 — the derived floor is always zero, on every load.**

`image_sizes_derive_floor` returns at the guard in `image_tables.cpp:486` because `ext_strucpp_get_located_vars` is still null. That pointer is assigned only in `symbols_init` (`image_tables.cpp:262`), and `symbols_init` has exactly one call site: `plc_state_manager.cpp:426`, inside `plc_cycle_thread` — a thread created at line **1125**, after the sizing block at 1076-1103. The second load does not escape it either: `image_tables_clear_null_pointers` sets the pointer back to null at `image_tables.cpp:1094` during unload.

So `image_sizes_take_max` is always taking the max of `image.conf` and a zero vector, and the `max(configured, derived)` design documented at `image_tables.h:95-115` never contributes. With no `image.conf`, `image_sizes_largest` returns 0, `image_tables_alloc` clamps to `IMAGE_MIN_ELEMENTS = 1`, and the runtime comes up with a one-element image for any program — every located address above index 0 silently rejected by the bounds check, no log. That is the acceptance criterion "removing the configuration file by hand still brings the runtime up, sized by the floor derived from the loaded program, rather than undersized", and it fails.

The happy path works, because the editor always emits `image.conf`. What does not work is the safety net built for when it is missing — an older editor, a device provisioned another way, the file removed by hand. The PR's reasoning is right that the floor needs the loaded `.so`; the gap is that `plugin_manager_load` dlopens it while `symbols_init` is what resolves the accessors, and that happens later, on another thread.

Worth noting how the testing missed it: `[image_tables] image allocated: 1 elements per table` in the boot log is correct for boot (`plc_main.c:160` passes 0), and it is also what every program load produces without `image.conf`. The check that would separate the two is loading a real program with located variables and reading back the capacity.

**BLOCKER 2 — `journal_init` deadlocks the mutex-fallback build on an allocation failure.** `journal_buffer.c:617` takes `g_journal_mutex`; if `force_map_alloc` fails, the `return -1` at 628 skips the `pthread_mutex_unlock` at 634. The mutex is never released, so every later `journal_add`, `journal_apply_and_clear` and `journal_is_initialized` blocks forever and the scan thread stops; `journal_cleanup` blocks too. The lock-free variant (416-455) takes no mutex in init, so its `return -1` at 439 is clean. Moving `force_map_alloc` above the lock, or unlocking before the return, fixes it.

**BLOCKER 3 — the journal write bound truncates and can disable every write.** `journal_buffer.c:155` compares `idx >= (uint16_t)g_buffer_ptrs.buffer_size`, and `buffer_size` is `int` (`plugin_types.h:233`). At capacity 65536 the cast yields 0 and every journal write is dropped with no diagnostic. The file's own new comment at lines 86-89 identifies exactly this wrap, which is why `g_force_size` was made `uint32_t` — lines 137, 329 and 356 compare correctly against it, and 155 is the only site left with the cast. 65536 is reachable: it is the ceiling of the uint16 `byte_index` in the STruC++ ABI.

**BLOCKER 4 — the twenty-four new tests never run in CI.** `.github/workflows/tests.yml:99` passes `--ignore=tests/pytest/plugins`, and both new files live there. The PR justifies the text-scanning contract test on the grounds that "pytest is the only suite CI runs in this repository, so it is the only guard that will actually fire" — at that path it does not fire. They do run under `scripts/run-pytest.sh:62`, which carries no ignore, which is why they passed locally. The ignore exists for pre-existing failures in those directories (the comment at lines 86-90 says so), so the fix is to move the two files to `tests/pytest/` rather than to drop the ignore.

---

**Required**

- `plc_state_manager.cpp:1084` — locks `image_tables_mutex()` before it is initialised. `init_recursive_pi_mutex` runs only inside `symbols_init` (`image_tables.cpp:313-321`), on the cycle thread, so on the first load this is a zero-filled `pthread_mutex_t`. It happens to work on glibc and is neither recursive nor PI there. Same root cause as blocker 1; initialising the image mutex once at process start would close both.
- `plc_main.c:160-165` — a failed boot allocation only logs and falls through to `plugin_driver_init`, whose return value is discarded. With capacity 0 the new guard refuses the args for every plugin, so the runtime boots with no plugins initialised and one log line. The criterion is log-and-stop.
- `image_tables.cpp:943` — `image_tables_alloc` calls `image_tables_free()` before it knows the new image fits. The header's all-or-nothing promise covers the new allocation but not the one it just destroyed: a failed re-allocation leaves capacity 0 and fourteen null tables while plugins still hold the old base pointers. Building into locals and publishing only on success would make the promise true.
- `plc_state_manager.cpp:1113` and `:1132` — the image is not released on either load rollback, and the boot image with no program is never released at shutdown.
- `plc_state_manager.cpp:1212` — `plugin_driver_stop` skips any plugin with `running == 0` (`plugin_driver.c:891`), and ethercat is deliberately initialised even when disabled (`plugin_driver.c:610-613`). An initialised-but-never-started plugin therefore keeps its by-value copy of the base pointers across `image_tables_free`. The comment at 1209-1211 is right about running plugins and does not cover these.
- `image_tables.cpp:474` — `strtol` with no `ERANGE` check and a silently truncating cast. The comment above promises a hand-edited value falls through to the derived floor, but an oversized one wins `image_sizes_take_max` and makes the allocation fail, taking the runtime to ERROR on a program it could size by itself.
- `webserver/image_config.py:138` — only `FileNotFoundError` is caught. The file arrives from an upload, so a non-UTF-8 `image.conf` raises `UnicodeDecodeError` and a directory entry named `image.conf` raises `IsADirectoryError`; both escape to `app.py:453/460`, which return `f"Unexpected error: {e}"` to the client. That is the `restapi.py:838` pattern, now reachable from an uploaded byte.
- `simple_modbus.py:897` — the shrink warning misses the case it exists for. A legacy-shape config (`max_coils`, `max_holding_registers`, …) leaves `requested.get("holding_registers", {})` empty so nothing is reported while lines 1018-1031 still clamp, and a config with no `buffer_mapping` returns at 898 while the 8192/1024 defaults at 1010-1013 are clamped in silence. The old-editor config is exactly what this is for.
- `simple_modbus.py:1105` — the clamp now admits up to 65536 per segment into dense pymodbus blocks: `qw + mw + 2*md + 4*ml` reaches 524288 `[0] * n` entries, well past the 65536 addresses a Modbus PDU can reach. It needs the user to configure large counts, but everything above 65536 is unaddressable by construction.
- `journal_buffer.c:100` — `force_map_alloc` overwrites `g_forced[t]` unconditionally, leaking fourteen rows on any `journal_init` not preceded by `journal_cleanup`.
- `journal_buffer.c:329` and `:356` — `journal_force_set` / `journal_force_clear` still return with no log when out of range; only the bound changed from 1024 to `g_force_size`. When the map was never allocated `g_force_size` is 0 and every force is dropped without a trace. Both run under `image_lock`, not on the lock-free producer path, so a rate-limited warning is affordable — and silent drops are what this change set out to remove.
- `image_tables.h:202` — `image_tables_alloc` and `image_tables_free` are the only image entry points with no documented locking contract, while bind/fill/clear immediately below all state "Caller must hold the image-tables mutex". The two call sites already disagree: `plc_state_manager.cpp:1084` locks, `plc_main.c:160` does not.
- `image_tables.h:187` and the PR body — the memory figure understates the cost by roughly three times. At 4096 elements on 64-bit, the three BOOL tables are `IEC_BOOL *[8]`, so 64 bytes per element, not 8: 786 KB. The other eleven add 360 KB, and the `temp_*` backing buffers are also sized at `elements` and are not counted at all, adding about 272 KB. That is roughly **1.36 MiB**, against the stated ~460 KB. The number matters because "the cost is bounded and small" is what carries the square-image decision over per-table sizing.

**Nits** — the first two `static_assert`s at `image_tables.cpp:57` compare each member against its own declared type, so they cannot catch the drift the comment claims (only the `sizeof(g_image) == 14 * sizeof(void *)` one does real work); the contract test skips any struct member not spelled `IEC_*`, which fails open; `MAX_BUFFER_SIZE` (`plugin_runtime_args.py:28`) and `MAX_TABLE_ELEMENTS` (`image_config.py`) are the same ABI fact written twice with only one of them pinned by a test; `image_config.py` copies `RUNTIME_ROOT`, the parse loop and the write-fsync-rename block from `retain_config.py` verbatim; `test_plugin_driver.c` is untouched, so the capacity-zero refusal and the alloc-before-init ordering have no C-side coverage at all; the four `log_*` fields in the ctypes mirror are declared non-variadic while the C side is `void (*)(const char *, ...)` (layout unaffected, prototype wrong).

---

**Cross-PR — this changes a finding I raised on editor#1093**

This PR independently found the same thing I flagged there and calls it "a case nobody had noticed, and it is the common one": `generateModbusSlaveConfig` materialises 1024 registers / 8192 coils into `modbus_slave.json` even when the user never opened the Modbus screen. The answer taken here is to clamp in the plugin (RSK05 as clamp-and-warn) rather than to change the editor.

That does resolve the contradiction on Runtime v4 — the client gets exception 02 instead of a plausible zero. Three things follow. The warning that makes the clamp visible is the `simple_modbus.py:897` item above, which misses precisely the old-editor shape. The two changes now have to land together, or the editor ships an `image.conf` nothing reads while `modbus_slave.json` keeps overstating. And bare metal has no Modbus plugin to clamp, so it is not covered there.

One thing for the Change Record rather than the code: `image_sizes_largest` collapses the fourteen figures into one, so the demand's "an area with no producers is sized to zero" — which the editor's risk assessment cites as its Data Minimization row — holds on bare metal only. On Runtime v4 every table is as large as the largest. That is a deliberate, well-argued consequence of CON06, but it is not what the assessment currently claims.

---

**On CI.** Three checks did run and pass here (Bootloader, Installer scripts, Webserver pytest). The Ceedling suite still has no gate, and blocker 4 means the pytest run that passed did not include the new tests.

core/src/plc_app/journal_buffer.c
```diff
@@ -71,25 +72,82 @@ static int journal_add(uint8_t type, uint16_t index, uint8_t bit, uint64_t value
...
+            g_force_size = 0;
+            return -1;
+        }
+    }
+    g_force_size  = elements;
+    g_force_count = 0;
+    return 0;
+}
+
+static void force_map_free(void)
```
marconetsf, line 230: **🔴 Major** · confidence 9/10 · use-after-free reachable from a plugin thread

Moving the forced-slot map from static storage to the heap gives it a lifetime it never had, and two things combine badly.

**First, the ordering inside this function is backwards.** The rows are freed and NULLed, and only afterwards are the guards cleared:

```c
for (int t = 0; t < JOURNAL_TYPE_COUNT; t++)
{
    free(g_forced[t]);
    g_forced[t] = NULL;
}
g_force_size  = 0;
g_force_count = 0;
```

`is_slot_forced` checks exactly those two variables before indexing:

```c
if (g_force_count == 0) return 0;
if (type >= JOURNAL_TYPE_COUNT || idx >= g_force_size) return 0;
...
return g_forced[type][idx] != 0;
```

A reader that passes both guards because they have not been cleared yet then dereferences a NULL row.

**Second, `journal_cleanup()` runs while plugin threads are still live.** In `plc_state_manager.cpp` it sits four lines above `plugin_driver_stop(plugin_driver)`, and the comment just above it already acknowledges plugins are alive at that point ("a plugin-backed store has to still be alive to answer"). `image_lock` is handed to every plugin as `args->image_lock` and calls `journal_apply_and_clear()`, so `is_slot_forced` runs on plugin-owned threads: EtherCAT's per-master `ecat_bus_thread` calls it at `ethercat_plugin.c:921`, and the s7comm server callback at `s7comm_plugin.cpp:1088`. `journal_cleanup` itself takes no lock.

So: stop the PLC with at least one variable forced (`g_force_count` is non-zero only then), and a plugin thread can be inside `is_slot_forced` reading a row that was just freed. Forcing a variable from the debugger or through OPC UA is an ordinary operation, so this is reachable rather than theoretical.

Both halves of the fix are worth doing:

1. In this function, set `g_force_count = 0; g_force_size = 0;` **before** the free loop. A reader that observes either guard cleared never indexes a row, which closes the NULL dereference on its own.
2. Move `journal_cleanup(); debug_write_journal_reset();` to **after** `plugin_driver_stop(plugin_driver)`. That is the only thing that guarantees no plugin thread is inside `image_lock()`, and it also closes the same-shaped race between the `memset(&g_buffer_ptrs, 0, ...)` in cleanup and `apply_write_raw` reading `g_buffer_ptrs.buffer_size`.

core/src/plc_app/plc_state_manager.cpp
```diff
@@ -1055,6 +1058,50 @@ extern "C" int load_plc_program(PluginManager *pm)
...
+            {
+                image_sizes_t configured;
+                image_sizes_t floor;
+                image_sizes_read_conf("./image.conf", &configured);
+                image_sizes_derive_floor(pm, &floor);
+                image_sizes_take_max(&configured, &floor);
+
+                pthread_mutex_t *itm = image_tables_mutex();
+                pthread_mutex_lock(itm);
+                const bool ok = image_tables_alloc(image_sizes_largest(&configured));
```
marconetsf: **🟡 Changes required** · confidence 8/10 · the image is gated on a plugin concern

This whole sizing and allocation block sits inside `if (plugin_driver)`, which opens at line 1034 and closes at 1133. The image is the **program's** storage, not the plugins' — it is where located variables live — so it should not be conditional on there being a plugin driver.

`plugin_driver` is the global set once in `plc_main.c`. If `plugin_driver_create()` returns NULL, `plc_main.c` simply skips its entire block: no error, no exit. The same is true one level down, where the boot-time `image_tables_alloc(0)` is nested inside `if (plugin_driver_load_config(...) == 0)` and the `else` only logs `"Failed to load plugin configuration"` and falls through.

A later START then reaches `load_plc_program` with `plugin_driver == NULL`, skips this block, and `g_capacity` stays 0 for the life of the process. Downstream: `image_tables_fill_null_pointers()` loops zero times, `journal_init` gets fourteen NULL bases and `buffer_size == 0`, `journal_add` rejects every write, and `threaded_image_read` returns zero for every located variable. The PLC scans and drives nothing, silently — which is the failure `plc_main.c` explicitly refuses to ship a few lines above the boot allocation ("a runtime that looks alive and drives nothing").

Suggest hoisting this block above line 1034, out of the `if (plugin_driver)`.

core/src/plc_app/plc_state_manager.cpp
```diff
@@ -1156,9 +1222,24 @@ extern "C" int unload_plc_program(PluginManager *pm)
...
+         * started or not, which is what actually ends the last reference. */
+        plugin_driver_cleanup_init(plugin_driver);
+
         pthread_mutex_t *itm = image_tables_mutex();
         pthread_mutex_lock(itm);
         image_tables_clear_null_pointers();
+        /* Released only after every plugin has been stopped AND de-initialised
+         * above. Both native plugins cache these pointers by value at init(),
+         * so freeing any earlier hands them memory that belongs to nobody. */
+        image_tables_free();
```
marconetsf: **🟡 Changes required** · confidence 8/10 · only half of the ordering is asserted

The acceptance criterion asks for the ordering to be *asserted*, not merely done, and the allocate-before-init half does that properly — `plugin_driver.c:1063` is a real runtime check with a comment explaining it is deliberately not `assert()` because it has to hold in the field.

The release-after-stop half has no equivalent. `image_tables_free()` is protected only by the comment above it. `plugin_driver_cleanup_init` already returns the count it cleaned and that return value is discarded here, so a refactor that moves this call above line 1234 gets no diagnostic at all — which is the exact class of regression the other half guards against. It is also not a purely hypothetical gap: `cleanup_init` skips any plugin whose `initialized` is 0, and on the native branch a plugin with no `cleanup` symbol keeps its by-value `g_runtime_args` copy with no way for the runtime to know.

Suggest using the returned count against a live count, or adding a `plugin_driver_any_initialized(driver)`, and refusing the free with a `log_error` when a plugin is still initialised — mirroring the guard in `generate_structured_args_with_driver`.

core/src/plc_app/journal_buffer.c
```diff
@@ -215,18 +337,27 @@ static void apply_entry(const journal_entry_t *entry)
...
-    if ((uint8_t)type >= JOURNAL_TYPE_COUNT || index >= JBUF_FORCE_SIZE) {
+    if ((uint8_t)type >= JOURNAL_TYPE_COUNT || index >= g_force_size)
+    {
+        /* Said out loud. A silent drop here is the exact defect this change
+         * set out to remove: someone forcing a high address from the debugger
+         * would watch nothing happen and have nothing to read. When the map
+         * was never allocated g_force_size is 0 and EVERY force lands here.
+         * Both force paths run under image_lock rather than on the lock-free
+         * producer path, so a log line is affordable. */
+        log_warn("Journal: force ignored, type %u index %u outside the image (%u slots)",
```
marconetsf: **🟡 Changes required** · confidence 8/10 · logging on the real-time path, under the image lock

Replacing the silent `return` was the right instinct — dropping a force with no diagnostic is the bug being fixed. The placement is the problem.

The comment argues "both force paths run under image_lock rather than on the lock-free producer path, so a log line is affordable". Being under `image_lock` is what makes it *not* affordable. The only caller is `apply_located`, reached from `debug_write_journal_drain()`, which runs inside `image_lock() ... image_unlock()` on `plc_cycle_thread` — the thread that called `set_realtime_priority()`.

`log_warn` takes `log_mutex`, a plain `PTHREAD_MUTEX_INITIALIZER` with no priority inheritance (unlike the image mutex, which is built through `init_recursive_pi_mutex` precisely because it needs it), and then does a blocking socket `write()`. So the SCHED_FIFO dispatcher can block behind a low-priority logging thread while holding the image lock, which stalls every plugin thread waiting on that lock. It is also unrate-limited: one line per offending entry, up to `DBGW_MAX_ENTRIES` = 128 per drain, and an OPC-UA client repeatedly writing one bad address reproduces it every tick.

Suggest keeping the guard but counting instead of logging — a `static unsigned g_force_oob_drops` reported from off the RT path, or the rate-limited pattern already used elsewhere in `plc_state_manager.cpp`. Better still, give `journal_force_set`/`journal_force_clear` an `int` return so `apply_located` can propagate the refusal back to the debugger or OPC-UA client, which is where the person who asked for the force is actually looking.

core/src/plc_app/plc_state_manager.cpp
```diff
@@ -453,21 +453,24 @@ void *plc_cycle_thread(void *arg)
...
+        .lint_input   = g_image.lint_input,
+        .lint_output  = g_image.lint_output,
+        .lint_memory  = g_image.lint_memory,
+        /* Follows the image: journal_buffer.c bounds every forced write
+         * against this, so a stale constant here would silently drop writes to
+         * the part of the image beyond it. */
+        .buffer_size  = (int)image_tables_capacity(),
         .image_mutex  = itm,
     };
     if (journal_init(&journal_ptrs) != 0)
```
marconetsf, line 490: **🟡 Changes required** · confidence 8/10 · a newly realistic failure is logged and ignored

```c
if (journal_init(&journal_ptrs) != 0)
{
    log_error("Failed to initialize journal buffer");
}
else
{
    log_info("Journal buffer initialized");
}
```

No return, no ERROR state. That was tolerable while `journal_init` could only fail on a NULL argument — a programming error caught at first boot. This PR adds `force_map_alloc` to both `journal_init` variants, so it now fails on allocation pressure at runtime, and that path leaves `g_initialized == false`.

With `g_initialized == false`, every `journal_add` returns -1 and `journal_apply_and_clear` returns immediately. That is not only "plugins cannot write": the program's own copy-out goes through the same journal (`image_tables.cpp` calls `journal_write_bool` / `journal_write_byte` there). So the runtime proceeds to `plugin_driver_start`, takes real-time priority, spawns the task threads and publishes RUNNING with every located output permanently dead, and one `log_error` line to say so.

That is the same class of silent failure this task set out to remove, one layer up. Suggest treating it as fatal and matching the image-allocation branch this PR itself added further down ("log and stop, never a partial image"): publish ERROR and return from `plc_cycle_thread`.

webserver/image_config.py
```diff
@@ -0,0 +1,296 @@
...
+                # there rather than being guessed at here.
+                count, _, unit = value.partition(" ")
+                units[key] = unit.strip()
+                try:
+                    sizes[key] = int(count)
+                except ValueError:
+                    # Left as a parse failure rather than an exception: a
+                    # garbled line should produce the same clear refusal as an
+                    # out-of-range one rather than a traceback from the parser.
+                    sizes[key] = -1
```
marconetsf: **🟡 Changes required** · confidence 9/10 · the refusal message describes the wrong mistake

`read_image_conf_file` stores `-1` as the sentinel for an unparseable count:

```python
try:
    sizes[key] = int(count)
except ValueError:
    sizes[key] = -1
```

`validate_table_count` then always receives an `int`, so its `except (TypeError, ValueError)` branch — the one that says "must be a whole number of {expected}" — is unreachable from the file path, and the `-1` falls into the `count < 0` branch instead. Running the real module:

```
int_output=4.5 words                  -> int_output cannot be negative (got -1).
int_output=<5000 digits> words        -> int_output cannot be negative (got -1).
```

The refusal is correct and nothing escapes, but the message is the entire point of validating here rather than in the core — this module's own docstring says refusing it here is the only place a person sees it. Someone who wrote `4.5` goes looking for a minus sign that is not there, and a completely different mistake (Python's int-from-string digit limit) produces the same sentence.

One-line fix: make the sentinel `None` instead of `-1`. `int(None)` raises `TypeError`, the existing handler fires, the message becomes "int_output must be a whole number of words.", and the negative branch goes back to meaning an actual negative.

marconetsf (commented):
Reviewed across three axes: the allocation lifecycle and memory safety, the journal and concurrency, and the Python side with the ABI mirror.

Severity on every point, here and inline: 🔴 Major (must fix before merge) · 🟡 Changes required (should be addressed) · 🟢 Nit (optional) · ❓ Question. Each carries a confidence out of ten.

**One 🔴 Major, five 🟡 Changes required and one ❓ Question**, all inline. The mechanical half of the change holds up well and I want to be specific about what I checked rather than leave it implied:

- **The ctypes mirror was measured, not just read.** I compiled a probe against the unmodified `plugin_types.h` and printed `offsetof` for all 39 members, then imported `shared.plugin_runtime_args` and printed every field's `.offset`. Identical, and `sizeof` is 552 on both sides. Nothing inserted, reordered, resized or removed, and both `10000` caps are gone with none left anywhere in the Python plugin tree. Given that nothing in CI checks this, it seemed worth doing properly.
- **All-or-nothing allocation verified line by line.** All 28 `calloc`s are issued, then checked as one conjunction; on failure all 28 locals are freed and the function returns without touching `g_image`, the `temp_*` buffers or `g_capacity`. No double free, no leak on the failure path.
- **The derived floor is right**: `byte_index + 1` on a `uint16_t`, no off-by-one, uniform across BOOL and the rest, an empty program yields zero, and nothing assumes a minimum. `max(configured, derived)` is genuinely a max with the units reconciled first.
- **The remaining `sizeof`s are correct** and the three `static_assert`s do still pin the shape.
- **All three new pytest files are actually run by CI** — none landed in one of the three ignored directories, and their docstrings show that was deliberate.
- No allocation on the scan path, nothing in the diff reachable from a signal handler, `JBUF_FORCE_SIZE` fully gone, no new `str(e)` leak, and the upload route is `@jwt_required()`.

Two things about the environment that are worth stating rather than leaving to inference. **The mutex journal variant is dead code**: `JOURNAL_FORCE_MUTEX` is defined nowhere in the repository and `ATOMIC_INT_LOCK_FREE == 2` on every supported target, so every build compiles the lock-free path. The asymmetries between the two variants are real but unobservable — and equally, nothing will catch a future divergence in that half. And **Ceedling still does not run in CI**, so the allocation work, whose whole risk lives in the second program load, has no automated gate. The three green jobs are real (`tests.yml` does have a `pull_request` trigger now) but they do not cover any of this.

core/src/plc_app/plc_state_manager.cpp
```diff
@@ -1156,9 +1222,24 @@ extern "C" int unload_plc_program(PluginManager *pm)
...
+        /* STOP IS NOT ENOUGH TO MAKE THE IMAGE FREEABLE, which is easy to miss.
+         *
+         * plugin_driver_stop skips any plugin whose `running` is 0, and a
+         * plugin can be initialised and never started -- ethercat is
+         * deliberately init'd even when disabled. Such a plugin still holds
+         * the by-value copy of the base pointers it took in init(), so
+         * stopping the running ones and freeing would leave it pointing at
+         * released memory. cleanup_init undoes init() for every plugin,
+         * started or not, which is what actually ends the last reference. */
+        plugin_driver_cleanup_init(plugin_driver);
```
marconetsf, line 1388: **❓ Question** · confidence 6/10 · raising this for the EtherCAT owner rather than as a blocker

Adding `plugin_driver_cleanup_init` here is necessary for the free below to be safe, and the reasoning in the comment above it is right. The side effect is that from the first STOP onwards every plugin is de-initialised, where previously they stayed initialised until the next load's `plugin_driver_update_config`.

`plugin_driver_execute_command` does not gate on that — it dispatches on `plugin->native_plugin && plugin->native_plugin->execute_command` with no `plugin->initialized` check, and `native_plugin` survives `cleanup_init` since only `teardown_plugin_instance` frees it. EtherCAT's `cleanup()` frees `g_masters`, nulls it and zeroes `g_master_count`, but leaves its cached `g_runtime_args` untouched.

So `PLUGIN_CMD` handlers issued while the PLC is stopped — which is exactly when the editor scans slaves and reads status — now run against torn-down state. I traced the dispatch but not every handler: `scan` and `list-interfaces` look bus-level and independent of `g_masters`, so I could not demonstrate a crash, but `status` and `diagnostics` iterate `g_masters` and will report nothing where they previously reported the configured masters. The comment a few lines above also notes EtherCAT is "deliberately init'd even when disabled", and this change ends that state at every stop.

Is that intended? If the editor-side commands do need the plugin initialised, either gate `plugin_driver_execute_command` on `plugin->initialized` and re-init on demand, or scope the de-init to the plugins that actually cached the image.

PR #198, opened 2026-09-17
fix(opcua): actually read and write STRING and WSTRING (DOPE-566, DOPE-646)

thiagoralves:
Fixes **DOPE-646**. No OPC-UA string value ever reached or left the PLC on Runtime v4.

## What was wrong

`_ctype_for()` returned `None` for STRING and WSTRING, so `debug_read_value` returned `None` and `debug_write_value` returned `False` **without ever calling the runtime** — on a justification that had gone stale:

```python
# STRING/WSTRING — not supported yet (Phase 4a in
# debug_dispatch.hpp explicitly stubs string reads).
```

strucpp has wired `read_string` / `write_string` / `read_wstring` / `write_wstring` into `type_ops[]` at tags 19/20 for some time. This is the **same stale assumption already corrected in the baremetal comments** (`modbus_debug.cpp` "string stub", `opcua_nodes.cpp` "handle_read returns 0 for them"); here it had survived as executable code rather than a comment.

It was invisible rather than loud, which is why it lasted:

- a read substituted a default and stamped it **`Good`**, so a client could not tell "the string is empty" from "this server cannot read strings"
- a write was answered **`Good`** after being dropped
- WSTRING had no entry in `map_plc_to_opcua_type`, so it fell through to the `VariantType.Variant` default and asyncua died serialising a nested Variant around an int (`'int' object has no attribute 'VariantType'`), taking the **whole response** with it as `BadInternalError`

**Proof the read was fabricated, not merely empty:** force `INSTANCE0.UA_MSG` to `"via debug"` over the runtime's own debug channel and read both ways at the same moment — debug said `'via debug'`, OPC-UA said `''`.

## What this changes

- `_decode_string` / `_encode_string` handle strucpp's `[count][payload]` wire form. `count` is in **characters** for STRING and **UTF-16 code units** for WSTRING, never bytes. `validate_payload` *refuses* an over-cap write rather than truncating, so the 126-unit truncation happens plugin-side.
- STRING is **UTF-8**, matching what the rest of the system already agrees on — the editor's debugger decodes this same wire form with its `len8-utf8` codec. Truncation walks back off any continuation byte so the byte budget never splits a character.
- WSTRING maps to **ByteString** (UTF-16LE code units, no transcoding) — the same choice the baremetal runtime makes, so a project behaves identically on both targets.
- a read that did not reach the PLC now returns `BadNoDataAvailable` instead of a default stamped `Good`.

A failed **write** is still reported `Good`; that one is documented in place rather than fixed. asyncua's `value_setter` cannot return a per-value StatusCode, and raising would fault the whole Write request including values that did land. It is now reached only by a genuinely refused write rather than by every string.

## Testing

Verified end to end in Docker running the **same program as a LOGO! 8.2** with only the target changed, every value cross-checked over the runtime's debug channel rather than trusting the OPC-UA client's reply:

| check | result |
| --- | --- |
| STRING round-trip / empty / non-ascii (`ünïcødé ✓`) | PASS |
| STRING 200 chars truncated to the 126 cap | PASS |
| WSTRING round-trip / empty | PASS |
| WSTRING 200 units truncated to the 126 cap | PASS |
| both strings distinct in one batched Read | PASS |
| OPC-UA write → debug read | `'batch-a'` / `'batch-b'` |
| debug force → OPC-UA read | `'forced via debug'` |

19 new unit tests cover the wire form, the cap, character-boundary truncation for 2/3/4-byte sequences, and the type mapping. Confirmed they bite: against the pre-fix code the debug-surface and mapping tests fail.

The 43 pre-existing failures in `tests/pytest/plugins/opcua` are **unchanged** by this PR — identical lists before and after. They are stale fixtures at `opcua.json` `format_version 0`.

### Worth knowing when testing

Runtime v4 **journals** debug writes and applies them at the no-task-running window, so an immediate read-back races the write. That is by design — it is what keeps plugin threads from racing the IEC task workers — and it is why baremetal (which applies immediately) behaves differently. My first test run reported false failures for exactly this reason.

https://claude.ai/code/session_01NegApcNt2SzRk3FDAs5T9p


Gustavohsdp (commented):
## Review — DOPE-566 / DOPE-646

The fix is sound and does what it claims. I verified the wire format against the source of truth rather than trusting the description, and validated locally because no workflow in this repo triggers on pull requests — a green page here proves nothing.

**Verdict: no blockers. Three `required` items before merge, the rest are nits.**

### What I confirmed

- Wire format matches `STruCpp/src/runtime/include/debug_dispatch.hpp:171-173,432-447`: `DEBUG_STRING_CAP = 126`, widths 127 / 253, and `validate_payload` refusing `count > CAP`. Plugin-side truncation was indeed mandatory.
- `_READ_BUFFER_SIZE = 256` covers the 253-byte WSTRING; the journal entry is `DBGW_MAX_BYTES = 256`, sized for the same width — no overflow, no truncation in transit.
- `ua.StatusCodes.BadNoDataAvailable` exists in the pinned `asyncua==1.1.8` (`0x80b10000`), so the new branch does not fall into the generic `except`.
- The 19 new tests pass, matching the count in the description.
- "Pre-existing failures unchanged" holds: I ran `tests/pytest/plugins/opcua` on `origin/development` and on this branch — the `FAILED` lists are byte-identical.
- DOPE-646 item 4 is done: no `Phase 4a` / `string stub` comment survives in the runtime.
- No runtime risk multiplier applies: this touches no plugin ABI, scan cycle, state machine, IPC, image table, webserver or packaging surface.

### Required

**1. `synchronization.py:362` — the comment understates the remaining hole.**

It says the surviving `Good` on a lost write is now reached only by "out-of-bounds coordinates, a read-only leaf, or a value that would not encode". On Runtime v4 neither of the first two can produce `False`:

- `plugin_debug_write` (`core/src/drivers/plugin_driver.c:150-151`) returns `0x7E` as soon as the write is **queued**, not when it is applied.
- `apply_global` (`core/src/plc_app/debug_write_journal.cpp:167-168`) discards the status `strucpp_debug_write` returns.

So an out-of-bounds leaf, a CONSTANT / read-only leaf and an over-cap payload are all reported `Good` **and are never logged either**. The only cases that still yield `False` are: no program loaded (`0x81`), journal queue full (`0x82`), and a Python-side encode failure. Please correct the comment to say that — as written it will tell the next reader the hole is narrower than it is.

**2. DOPE-646 suggested fix 3 is half-done, with no follow-up.**

"Stop returning `Good` from a write that returned `False`" is not implemented. The reasoning in the comment is good and I am not asking you to force it into this PR, but combined with item 1 this leaves silent data loss on a protocol write with nothing tracking it. Please open a follow-up issue and link it from that comment block, so the limitation has an owner.

**3. `synchronization.py:468-478` — the same pattern survives one function away.**

`_push_array_node` still substitutes `_get_default_value` for a failed element and pushes the whole array with `StatusCodes.Good` (`_push_value:453`). That is exactly what this PR fixed in the read callback. A subscribed STRING array can still deliver a fabricated empty stamped `Good`.

**4. PR body links neither the requirements document nor the cybersecurity risk assessment.**

This targets `development`. If this is a bug fix with no requirements document and no assessment required, that is fine — but it has to be stated explicitly in one line rather than left silent.

### Nits

- **`opcua_memory.py:61`** — "`count` is in CHARACTERS for STRING" is wrong: it is **bytes**. `IECString` stores `char data_[MaxLen + 1]` with `length_` counting bytes (`STruCpp/src/runtime/include/iec_string.hpp:340-341`), and `_truncate_utf8:133` in this same file already says "The cap is a BYTE budget". The two comments contradict each other. This matters to users: the cap is ~63 accented characters or ~42 emoji, not 126 characters.
- **`tests/pytest/plugins/opcua/test_string_types.py:20`** — the docstring lists "a failed read reported Bad, not as a default stamped Good" under "Covered here", but no test touches `synchronization.py` or any StatusCode. The one behaviour change in `synchronization.py` has no unit test.
- **`opcua_memory.py:115-118`** — WSTRING truncation (`count = min(len(raw) // 2, CAP)`) can split a surrogate pair, while the STRING path goes to real trouble not to split a UTF-8 sequence. Asymmetric care.
- **`opcua_memory.py:265`** — `debug_force_value` still returns `False` for strings. Read and write got a path, force did not. Nothing calls it today, but it is the same silence that produced this bug.
- **`opcua_memory.py:86,91`** — `log_warn` inside the read callback, which runs per variable per client Read. A persistently malformed leaf floods the log at client poll rate.
- **Duplication** — the per-type default table now lives in four places: `address_space.py:44`, `synchronization.py:486`, `opcua_utils.py:264` and `opcua_utils.py:405`. This diff adds the WSTRING row to three of them.
- **`test_string_types.py:126`** — `test_wstring_rejects_an_odd_byte_count` does not test a rejection: `_encode_string` drops the trailing byte and returns a valid one-unit payload. The name says reject, the behaviour truncates.
- **`test_string_types.py:27-31`** — `conftest.py:24-29` already inserts the same directories into `sys.path`.
- **Branch name** — `fix/opcua-string-wstring-support` departs from the `RTOP-<n>-<slug>` convention in `CLAUDE.md`, and the demand lives in DOPE while this repo's tracker is RTOP. A question for the team, not a fault of this PR.

### Risk: Medium (4)

+1 size (361 lines, 4 files), +1 client-visible OPC-UA contract change (WSTRING as ByteString), +1 status-code change on a network surface affecting *all* datatypes, -1 for test coverage.

thiagoralves:
## Review addressed — `4a03d60`

All four `required` items and every nit, plus one correction to the review itself.

### Required

**1. `synchronization.py:362` understated the hole — fixed.** Verified in the C rather than taking it: `plugin_driver.c:150` returns `0x7E` as soon as the write is *queued*, and `debug_write_journal.cpp:167` calls `ext_strucpp_debug_write(...)` and **discards** the status. So the comment named exactly the three cases that *cannot* produce `False`. It now says what is actually left — no program loaded (`0x81`), journal queue full (`0x82`), a Python-side encode failure — and that an out-of-bounds leaf, a `CONSTANT` leaf and an over-cap payload are reported `Good` *and never logged*.

**2. Follow-up opened — DOPE-647**, linked to DOPE-646, and referenced from that comment block. Carrying the applied status back out of the journal is a change to the journal contract, not to this plugin, so it has its own ticket rather than being forced in here.

**3. `_push_array_node` — fixed.** It now marks the push `BadNoDataAvailable` when any element fails to read. Your scoping was right: the scalar path needs nothing, because its caller `continue`s on a failed read and leaves the client's last good value in place.

**4. PR body** — links added.

### Nits, all fixed

`count` is in **bytes** for STRING (the comment contradicted `_truncate_utf8` in the same file, and the cap is ~63 accented characters or ~31 emoji, not 126 of either) · the test docstring no longer claims StatusCode coverage it did not have — `test_read_status.py` now covers it for real, 6 cases · WSTRING truncation no longer splits a surrogate pair · `debug_force_value` handles strings through the same wire form as a write · the read-callback warnings are once-per-problem instead of once per variable per Read · the four per-type default tables are two, which surfaced a real drift (`address_space` seeded WSTRING as `""` where everything else used `b""`, and WSTRING is served as a ByteString) · `test_wstring_rejects_an_odd_byte_count` renamed for what it asserts · the duplicated `sys.path` block is gone.

### One correction

> no workflow in this repo triggers on pull requests — a green page here proves nothing

Three do, and they are green: Bootloader (Go), Webserver (pytest), Installer scripts. **Your conclusion survives for a worse reason**: `tests.yml:98` runs `pytest tests/pytest --ignore=tests/pytest/plugins`, and this PR's tests live under exactly that path. The tick was real and vacuous — and it *looked* like coverage.

Fixed: a second step names the plugin suites that pass on a clean checkout and runs them. CI now reports `154 passed` for the main step and `26 passed` for this PR's own tests. The blanket ignore stays, since those suites are red on the base branch and a gate that can never pass is one everyone learns to ignore.

### Verified on a real Runtime v4

Deployed to a v4 container and driven over OPC-UA. The pre-existing suite still passes 8/8 on both the plain and `Basic256Sha256/SignAndEncrypt` endpoints, and the string work is confirmed live: 80 accented characters truncate to **63 — exactly 126 bytes**, and a truncated WSTRING comes back 250 bytes (125 code units) and decodes cleanly, so the pair was dropped whole. Pre-existing plugin-suite failures are unchanged against the base branch.

PR #200, opened 2026-09-19
OPC-UA: honor explicit anonymous session role (default viewer)

thiagoralves:
### Problem
An anonymous OPC-UA session has no identity, so the plugin **inferred** its role: with no users configured it granted anonymous **engineer/Admin** (full write), silently overriding the per-variable Viewer/Operator = read-only permissions an administrator had configured. The intent was guessed, and "anonymous can write" was the silent default. (Found during 4.3.0 stress testing: an anonymous client could write a variable whose Viewer/Operator permission was read-only.)

### Change
`_authenticate_anonymous` now takes the role from the security profile's `anonymous_role` field instead of inferring it. The editor defaults that field to least-privilege `viewer`, so anonymous is read-only out of the box; allowing unauthenticated writes is a deliberate choice (raise the role to operator/engineer). Unknown values fall back to viewer. The config model parses `anonymous_role`, defaulting to `viewer` for projects authored before the field.

Per-variable permissions still apply on top of the role.

### Compatibility
Deliberate security tightening: an old project's anonymous sessions become read-only until an admin raises the role. Anyone who relied on anonymous writes sets the anonymous role to engineer in the editor.

### Pairs with
- openplc-web #771 / openplc-editor #1129 (the "Anonymous session role" dropdown + `anonymous_role` in the generated config).

### Verification (end-to-end on a vPLC)
- anonymous_role = viewer: external anonymous client write to a viewer=read-only variable is **denied** (`DENY write for user anonymous (role: viewer)`), value unchanged.
- anonymous_role = engineer: the same write is **accepted and persisted**.

https://claude.ai/code/session_01NegApcNt2SzRk3FDAs5T9p


thiagoralves (commented):
Code review of the anonymous-role change. The core direction is right (explicit beats inferred, and this also un-breaks `test_anonymous_allowed_when_profile_supports`, which has been failing since 6f98e83 because it asserts `viewer` with zero users configured). Three correctness notes below, all about how the new role string is read and validated.

core/src/drivers/plugins/python/opcua/user_manager.py
```diff
@@ -557,16 +558,22 @@ def _authenticate_anonymous(self, profile: Any) -> tuple[Optional[User], Optiona
...
-            # writes work without having to set up users.
-            openplc_role = "engineer"
-            asyncua_role = UserRole.Admin
-        else:
-            # Users configured — anonymous is read-only viewer.
+        # Explicit, config-driven role (defaults to viewer). Map to the asyncua
+        # role for operation-level checks; per-variable enforcement uses the
+        # OpenPLC role string.
+        openplc_role = getattr(profile, "anonymous_role", "viewer") or "viewer"
+        if openplc_role not in self.ROLE_MAPPING:
```
thiagoralves: **The role string is matched case-/whitespace-sensitively here, unlike everywhere else roles are compared.**

`ROLE_MAPPING` keys are lowercase, and this is an exact `in` test on the raw config value. Every other role consumer normalizes first: `callbacks.py:_normalize_role()` does `role.lower()` before comparing, which is why a user entry of `"role": "Engineer"` still enforces engineer permissions per variable. Note also that the sibling fields in the very same profile object are capitalized by convention (`auth_methods: ["Anonymous", "Username"]`), so a config (hand-edited, or an editor build that mirrors that convention) carrying `"anonymous_role": "Engineer"` is entirely plausible.

Scenario: admin sets the anonymous role to Engineer, config lands as `"Engineer"`. Instead of Admin/engineer, the session silently degrades to viewer with only a `log_warn` ("Unknown anonymous_role"), and every anonymous write is denied — the admin's explicit choice is discarded. It fails closed, but the symptom ("I set engineer and writes are still denied") is hard to diagnose.

Suggest normalizing once: `openplc_role = str(getattr(profile, "anonymous_role", "viewer") or "viewer").strip().lower()`.

core/src/drivers/plugins/python/opcua/user_manager.py
```diff
@@ -557,16 +558,22 @@ def _authenticate_anonymous(self, profile: Any) -> tuple[Optional[User], Optiona
...
-            # No user model configured — give anonymous full role so
-            # writes work without having to set up users.
-            openplc_role = "engineer"
-            asyncua_role = UserRole.Admin
-        else:
-            # Users configured — anonymous is read-only viewer.
+        # Explicit, config-driven role (defaults to viewer). Map to the asyncua
+        # role for operation-level checks; per-variable enforcement uses the
+        # OpenPLC role string.
+        openplc_role = getattr(profile, "anonymous_role", "viewer") or "viewer"
```
thiagoralves: **`profile` here is not necessarily the profile the client actually connected through, so a per-profile role can be applied to the wrong endpoint.**

`get_user()` obtains `profile` from `_find_profile_by_auth_method("Anonymous")` (line ~700), which returns the *first enabled* profile whose `auth_methods` contains `"Anonymous"` — the security policy/mode of the client's actual endpoint is not available at this layer (as the `get_user` docstring notes). That was harmless while the anonymous role was derived from the global user list; now that the role is a per-profile field, the field promises a per-endpoint decision the lookup cannot honor.

Scenario: an admin configures `insecure` (None/None, Anonymous, `anonymous_role: viewer`) plus `secure_anon` (Basic256Sha256/SignAndEncrypt, Anonymous, `anonymous_role: engineer`), intending anonymous writes only over the encrypted endpoint. Because `insecure` is listed first, the encrypted anonymous client gets viewer and cannot write; reorder the list (or disable `insecure`) and a plaintext anonymous client gets engineer/Admin and can write. Either way the role follows list order, not the endpoint.

At minimum worth an explicit constraint: log a warning at load if more than one enabled profile offers `Anonymous` with differing `anonymous_role` values, and document that the first one wins.

core/src/drivers/plugins/python/shared/plugin_config_decode/opcua_config_model.py
```diff
@@ -66,12 +71,16 @@ def from_dict(cls, data: Dict[str, Any]) -> 'SecurityProfile':
         except KeyError as e:
             raise ValueError(f"Missing required field in security profile: {e}")
 
+        # Optional; default to viewer for backward compatibility.
+        anonymous_role = data.get("anonymous_role", "viewer") or "viewer"
```
thiagoralves: **Parse time is where an invalid `anonymous_role` should be caught, not authentication time.**

This accepts any value (including a non-string — `"anonymous_role": 3` passes through, then silently misses the `ROLE_MAPPING` lookup in `user_manager`). The only feedback is a `log_warn` emitted inside `_authenticate_anonymous`, i.e. once per anonymous session: a typo'd role produces no error at server start (when the admin would see it), and then one warning line per client connect afterwards — noisy on a server polled by several clients, and easy to miss entirely in between.

Other enum-ish fields in this module are constrained at parse time (see `VALID_DATATYPES`). Suggest normalizing and validating here — `anonymous_role = str(data.get("anonymous_role") or "viewer").strip().lower()`, falling back to `"viewer"` with a single load-time warning if it is not one of viewer/operator/engineer — which also makes the `user_manager` side a plain lookup.

Separately: `docs/opcua/OPCUA_CONFIGURATION_GUIDE.md` documents every security-profile field in a table (around line 114) and is not updated here; a new security-relevant field that flips the default behavior of existing projects should be in that table and in the sample profile JSON above it.

core/src/drivers/plugins/python/opcua/user_manager.py
thiagoralves: Fixed in cd139ac. Added a shared `normalize_role()` in opcua_config_model and `_authenticate_anonymous` now runs the role through it (strip + lowercase, same normalization callbacks applies), so "Engineer"/" engineer " map correctly instead of degrading to viewer. Also normalized+validated at parse (see the config_model thread), so the value reaching here is already clean. Tests: test_anonymous_role_is_normalized_case_and_whitespace, test_anonymous_explicit_engineer_role.

core/src/drivers/plugins/python/opcua/user_manager.py
thiagoralves: Agreed the real fix is in the editor. Done: openplc-web#771 / openplc-editor#1129 now reject a config with more than one enabled Anonymous profile (validateOpcUaConfig). Runtime is belt-and-suspenders per your steer — cd139ac adds a load-time warning in UserManager.__init__ when >1 enabled profile offers Anonymous, names the winner, and keeps the current first-in-list-order behaviour. Test: test_multiple_anonymous_profiles_warns_and_first_wins.

core/src/drivers/plugins/python/shared/plugin_config_decode/opcua_config_model.py
thiagoralves: Fixed in cd139ac. SecurityProfile.from_dict now normalizes (strip+lower) and validates anonymous_role at parse, raising ValueError with a clear message if it is not viewer/operator/engineer — consistent with VALID_DATATYPES, surfaced once at server start instead of a per-session log_warn. A non-string like 3 is rejected too. Docs guide updated: field-table row, sample JSON, and an Anonymous-role note incl. the one-Anonymous-profile guidance. Tests in test_security_profile_anonymous_role.py.

